PR #33, opened 2018-06-29
Declarative over imperative

TheMallen:
After chatting with @lemonmade  about this I decided to propose this as a foundational principle. I think we've informally been pushing it more and more as Web has grown and it will be helpful to have it written down somewhere.

I'm not sure who else to ping on this tbh.

### DIFF
--- a/Principles/6 - Declarative over imperative.md
+++ b/Principles/6 - Declarative over imperative.md
@@ -1,0 +1,27 @@
+# Declarative over imperative APIs
+
+Imperative APIs are often a natural choice for developers who have spent a lot of time in C-style languages. The primitives of most popular programming languages are very easy to get started with writing highly imperative code. However, building imperative APIs can lead to much more error prone systems, more difficult to read code, and it being harder to discern the intention from the implementation for a given piece of logic.
+
+We prefer to write code that exposes a declarative API than an imperative API. A declarative API will let you define what results you would like to see, rather than the way you want them to happen. This type of API leads to simpler code with a higher degree of readability, maintainability, and composability. Where imperative code must exist we often counsel using more declarative and immutable styles within it.
+
+Below are a few examples of decisions that demonstrate this principle, some at the highest level of technology choice, and some at the smallest level of individual function implementations:
+
+* We prefer foundational technologies that are declarative, such as JSX and GraphQL, rather than imperative DOM manipulation and REST calls. On the build tooling side we choose the declarative Webpack over more imperative tools like Gulp.
+
+* We prefer to render a component that encapsulates a series of imperative DOM actions, such as `<EventListener />` or `<ScrollLock />`, than to expose classes or functions that one would call directly.
+
+* We prefer to use or build declarative APIs, such as `<FormState />` or `<JobPoller />` on top of complex state heavy operations such as managing form state, handling graphql requests and caching, and tracking long running jobs.
+
+* We prefer to expose server modules as declarative middleware functions rather than as stateful imperative classes.
+
+* We prefer to write stateless functional components wherever it is feasible.
+
+* We prefer to express program logic through a series of `const` assignments rather than mutating a single variable.
+
+* We prefer to express list and object manipulation, through functional constructs (map / reduce) than to build up a new list through a for loop.
+
+## Conflicts
+
+Sometimes it's correct to choose the more imperative path on a problem. This usually happens when the declarative path to a solution negatively effects user experience. In these cases the best path forward is to, as always, think of the user first. We should always write the code in the way that best addresses the problems our users face.
+
+ Usually these spots are relatively isolated and can be abstracted behind a declarative API for future use, or they speak to a missing component or tool that would let us accomplish them more gracefully in the long term. This doesn't always have to be addressed immediately, or even at all,but it can be a good way to look at particularly ugly pieces of code.
--- a/Principles/6 - Declarative over imperative.md
+++ b/Principles/6 - Declarative over imperative.md
@@ -1,27 +1,31 @@
-# Declarative over imperative APIs
+# Declarative over imperative
 
-Imperative APIs are often a natural choice for developers who have spent a lot of time in C-style languages. The primitives of most popular programming languages are very easy to get started with writing highly imperative code. However, building imperative APIs can lead to much more error prone systems, more difficult to read code, and it being harder to discern the intention from the implementation for a given piece of logic.
+Imperative code is often a natural choice for developers who have spent a lot of time in C-style languages. The primitives of most programming languages make it easy to write imperative code. However, imperative code can lead to more error prone systems and more difficult to read code. Imperative code can make it difficult to discern the intention from the implementation for a given piece of logic.
 
-We prefer to write code that exposes a declarative API than an imperative API. A declarative API will let you define what results you would like to see, rather than the way you want them to happen. This type of API leads to simpler code with a higher degree of readability, maintainability, and composability. Where imperative code must exist we often counsel using more declarative and immutable styles within it.
+We prefer to write declarative code. This style allows us to define the relationships between data and UI to describe what results we would like to see, rather than the way we want them to happen. Declarative code is generally more concise, readable and composable.
 
-Below are a few examples of decisions that demonstrate this principle, some at the highest level of technology choice, and some at the smallest level of individual function implementations:
+Below are a few examples of decisions that demonstrate this principle:
 
-* We prefer foundational technologies that are declarative, such as JSX and GraphQL, rather than imperative DOM manipulation and REST calls. On the build tooling side we choose the declarative Webpack over more imperative tools like Gulp.
+* We chose React over more traditional imperative DOM manipulation, or technologies like web components, for its declarative API.
+
+* We chose GraphQL over REST for its declarative query language and the declarative components provided by its client libraries.
 
 * We prefer to render a component that encapsulates a series of imperative DOM actions, such as `<EventListener />` or `<ScrollLock />`, than to expose classes or functions that one would call directly.
 
-* We prefer to use or build declarative APIs, such as `<FormState />` or `<JobPoller />` on top of complex state heavy operations such as managing form state, handling graphql requests and caching, and tracking long running jobs.
+* We prefer to use declarative APIs on top of complex state heavy operations. Examples include managing form state with `<FormState />`, handling graphql requests and caching with `apollo-client`, and tracking long running jobs with `<JobPoller />`.
 
-* We prefer to expose server modules as declarative middleware functions rather than as stateful imperative classes.
+* We prefer to build declarative components to express HTTP response behavior than to write imperative server code. For example; we built `<Redirect />` to replace imperative calls to `location.assign`.
 
-* We prefer to write stateless functional components wherever it is feasible.
+* We prefer to build components as pure functions when we can avoid writing a class. Classes can often result in more imperative code since they encourage manipulation of instance state.
+
+* We prefer declarative decorators for enhancing the behavior of class methods, such as `@autobind` or `@memoize`, over imperatively overwriting fields with wrapped versions or writing memoization logic by hand.
 
 * We prefer to express program logic through a series of `const` assignments rather than mutating a single variable.
 
-* We prefer to express list and object manipulation, through functional constructs (map / reduce) than to build up a new list through a for loop.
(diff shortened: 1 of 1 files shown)


Principles/6 - Declarative over imperative.md
```diff
@@ -0,0 +1,27 @@
+# Declarative over imperative
+
+Imperative code is often a natural choice for developers who have spent a lot of time in C-style languages. The primitives of many popular programming languages are very easy to get started with writing highly imperative code. However, building imperative APIs can lead to much more error prone systems, more difficult to read code, and it being harder to discern the intention from the implementation for a given piece of logic.
+
+We prefer to write code that exposes a declarative API than an imperative API. A declarative API will let you define what results you would like to see, rather than the way you want them to happen. This type of API leads to simpler code with a higher degree of readability, maintainability, and composability. Where imperative code must exist we often counsel using more declarative and immutable styles within it.
+
+Below are a few examples of decisions that demonstrate this principle, some at the highest level of technology choice, and some at the smallest level of individual function implementations:
```
TheMallen: other choices I thought about mentioning but didn't because they had too much overlap or seemed not super relevant

- imports over require. Import is declarative, require is imperative
-  typescript instead of runtime checks. Typescript is declarative, runtime checks are imperative
- SCSS is a declarative choice vs CSS in JS or some other crazy thing, seemed too trivial but it *is* a declarative choice
- how much we hate graphql cache manipulation - we don't have a solution yet but it's one of the worst parts of our codebase. Once we have a solution (which will assumedly be declarative) it should go here.

Principles/6 - Declarative over imperative.md
```diff
@@ -0,0 +1,27 @@
+# Declarative over imperative
+
+Imperative code is often a natural choice for developers who have spent a lot of time in C-style languages. The primitives of many popular programming languages are very easy to get started with writing highly imperative code. However, building imperative APIs can lead to much more error prone systems, more difficult to read code, and it being harder to discern the intention from the implementation for a given piece of logic.
```
lemonmade: second sentence reads better as “the primitives of most programming languages make it easy to write imperative code”

lemonmade: Switching to talking about imperative APIs feels weird, just stick to “code”

lemonmade: Also avoid the “much”s and similar, the sentences are plenty strong without them

Principles/6 - Declarative over imperative.md
```diff
@@ -0,0 +1,27 @@
...
+
+We prefer to write code that exposes a declarative API than an imperative API. A declarative API will let you define what results you would like to see, rather than the way you want them to happen. This type of API leads to simpler code with a higher degree of readability, maintainability, and composability. Where imperative code must exist we often counsel using more declarative and immutable styles within it.
```
lemonmade: First sentence is missing a word/ doesn’t read wel. Would also like to get away from the API part

lemonmade: Avoid the “we”, maybe just, “a developer should attempt to add a more declarative wrapper”?

Principles/6 - Declarative over imperative.md
lemonmade: I don’t feel we have convinced people why imperative code is more error prone/ harder to read

Principles/6 - Declarative over imperative.md
lemonmade: I would just ditch the entire part of this paragraph starting with “, some”

Principles/6 - Declarative over imperative.md
```diff
@@ -0,0 +1,27 @@
...
+
+Below are a few examples of decisions that demonstrate this principle, some at the highest level of technology choice, and some at the smallest level of individual function implementations:
+
+* We prefer foundational technologies that are declarative, such as JSX and GraphQL, rather than imperative DOM manipulation and REST calls. On the build tooling side we choose the declarative Webpack over more imperative tools like Gulp.
```
lemonmade: I would separate some of these out, we chose React over other librar/ technologyy X(web components maybe?) because of the declarative API. We chose GraphQL over REST.

lemonmade: The build tool example is not a particularly strong one so I would ditch it

Principles/6 - Declarative over imperative.md
```diff
@@ -0,0 +1,27 @@
...
+
+* We prefer to render a component that encapsulates a series of imperative DOM actions, such as `<EventListener />` or `<ScrollLock />`, than to expose classes or functions that one would call directly.
+
+* We prefer to use or build declarative APIs, such as `<FormState />` or `<JobPoller />` on top of complex state heavy operations such as managing form state, handling graphql requests and caching, and tracking long running jobs.
```
lemonmade: I would lead with the thing we prefer, then the examples.

Principles/6 - Declarative over imperative.md
```diff
@@ -0,0 +1,27 @@
...
+
+* We prefer to expose server modules as declarative middleware functions rather than as stateful imperative classes.
+
+* We prefer to write stateless functional components wherever it is feasible.
+
+* We prefer to express program logic through a series of `const` assignments rather than mutating a single variable.
+
+* We prefer to express list and object manipulation, through functional constructs (map / reduce) than to build up a new list through a for loop.
```
lemonmade: No space before /

Principles/6 - Declarative over imperative.md
```diff
@@ -0,0 +1,27 @@
...
+
+* We prefer foundational technologies that are declarative, such as JSX and GraphQL, rather than imperative DOM manipulation and REST calls. On the build tooling side we choose the declarative Webpack over more imperative tools like Gulp.
+
+* We prefer to render a component that encapsulates a series of imperative DOM actions, such as `<EventListener />` or `<ScrollLock />`, than to expose classes or functions that one would call directly.
+
+* We prefer to use or build declarative APIs, such as `<FormState />` or `<JobPoller />` on top of complex state heavy operations such as managing form state, handling graphql requests and caching, and tracking long running jobs.
+
+* We prefer to expose server modules as declarative middleware functions rather than as stateful imperative classes.
+
+* We prefer to write stateless functional components wherever it is feasible.
```
lemonmade: Not sure this is adequately explained for a beginner to understand what this even is, let alone how it demonstrates the principle

Principles/6 - Declarative over imperative.md
```diff
@@ -0,0 +1,27 @@
...
+
+Below are a few examples of decisions that demonstrate this principle, some at the highest level of technology choice, and some at the smallest level of individual function implementations:
+
+* We prefer foundational technologies that are declarative, such as JSX and GraphQL, rather than imperative DOM manipulation and REST calls. On the build tooling side we choose the declarative Webpack over more imperative tools like Gulp.
+
+* We prefer to render a component that encapsulates a series of imperative DOM actions, such as `<EventListener />` or `<ScrollLock />`, than to expose classes or functions that one would call directly.
+
+* We prefer to use or build declarative APIs, such as `<FormState />` or `<JobPoller />` on top of complex state heavy operations such as managing form state, handling graphql requests and caching, and tracking long running jobs.
+
+* We prefer to expose server modules as declarative middleware functions rather than as stateful imperative classes.
```
lemonmade: Feels like a weaker example given that we do end up using things that work this way (router), and for individual middlewares there is no choice but to do this. I would rather bring in how we prefer exposing HTTP response values through components in our application (like Redirect), rather than relying on having any server code at all, which by its nature is imperative

Principles/6 - Declarative over imperative.md
TheMallen: Is it our job to convince in these? I don't know how we could do that without code examples.

Principles/6 - Declarative over imperative.md
TheMallen: Let's make a JSX based server ;)

Principles/6 - Declarative over imperative.md
```diff
@@ -0,0 +1,31 @@
+# Declarative over imperative
+
+Imperative code is often a natural choice for developers who have spent a lot of time in C-style languages. The primitives of most programming languages make it easy to write imperative code. However, building imperative APIs can lead to more error prone systems, more difficult to read code, and it being harder to discern the intention from the implementation for a given piece of logic.
+
+We prefer to write declarative code. This style allows us to define the relationships between data and UI and what results we would like to see, rather than the way we want them to happen. Where imperative code must exist a developer should attempt to add a more declarative wrapper encapsulating it as an implementation detail.
```
lemonmade: data and UI and... reads weird

lemonmade: Last sentence is maybe an application of the decision?

Principles/6 - Declarative over imperative.md
```diff
@@ -0,0 +1,31 @@
+# Declarative over imperative
+
+Imperative code is often a natural choice for developers who have spent a lot of time in C-style languages. The primitives of most programming languages make it easy to write imperative code. However, building imperative APIs can lead to more error prone systems, more difficult to read code, and it being harder to discern the intention from the implementation for a given piece of logic.
```
lemonmade: Maybe “more difficulty” over “harder to” in the last sentence for flow

Principles/6 - Declarative over imperative.md
```diff
@@ -0,0 +1,31 @@
...
+
+We prefer to write declarative code. This style allows us to define the relationships between data and UI and what results we would like to see, rather than the way we want them to happen. Where imperative code must exist a developer should attempt to add a more declarative wrapper encapsulating it as an implementation detail.
+
+Below are a few examples of decisions that demonstrate this principle:
+
+* We chose React over more traditional imperative DOM manipulation, or technologies like web components, for it's declarative API.
```
lemonmade: Wrong it’s

Principles/6 - Declarative over imperative.md
```diff
@@ -0,0 +1,31 @@
...
+
+* We chose GraphQL over REST for it's declarative query language and the declarative components provided by it's client libraries.
```
lemonmade: More its

Principles/6 - Declarative over imperative.md
```diff
@@ -0,0 +1,31 @@
...
+
+* We prefer to render a component that encapsulates a series of imperative DOM actions, such as `<EventListener />` or `<ScrollLock />`, than to expose classes or functions that one would call directly.
+
+* We prefer to use or build declarative APIs on top of complex state heavy operations. Examples include managing form state with `<FormState />`, handling graphql requests and caching with `apollo-client`, and tracking long running jobs with `<JobPoller />`.
```
lemonmade: Remove “or build”

Principles/6 - Declarative over imperative.md
```diff
@@ -0,0 +1,31 @@
...
+
+* We prefer to build declarative components to express HTTP response behavior than to write imperative server code. For example; we built `<Redirect />` to replace imperative calls to `location.assign`.
+
+* We prefer to build components as pure functions when we can avoid writing a class. Classes can often result in more imperative code since they encourage manipulation of instance state.
+
+* We prefer declarative decorators for enhancing the behavior of class methods, such as `@autobind` or `@memoize`, over imperatively overwriting fields with wrapped versions or writing memoization logic by hand.
```
lemonmade: And for ease of composability, which is typically a result of declarative code

Principles/6 - Declarative over imperative.md
```diff
@@ -0,0 +1,31 @@
...
+
+* We prefer to express program logic through a series of `const` assignments rather than mutating a single variable.
+
+* We prefer to express list and object manipulation, through functional constructs (map/filter/reduce) than to build up a new list through a for loop.
```
lemonmade: Doesn't read write

Principles/6 - Declarative over imperative.md
```diff
@@ -0,0 +1,31 @@
...
+
+## Conflicts
+
```
lemonmade: I’m not sure this section adds that much value

TheMallen: I think it's important to highlight tensions between principles and times when we might want to make an exception

Principles/6 - Declarative over imperative.md
TheMallen: someday pipe operator will make it just as easy using function call syntax ;p

PR #35, opened 2018-07-10
Add docs about decision around, and alternatives to, testID

lemonmade:
Does what it says on the tin. I wanted to get this down into words so people understand why we are mostly opposed to the use of this attribute, provide some historical context for why we have so many in the existing Web app, and give people alternatives for the common patterns I have seen.

cc/ @Shopify/web-foundation

### DIFF
--- a/Best practices/Testing.md
+++ b/Best practices/Testing.md
@@ -21,6 +21,8 @@ Below are some specific applications of this principle:
 
 * Do not over-mock. Mocking can be a useful tool for short-circuiting potentially expensive function calls or for more easily simulating a particular environment. However, overuse of mocks reduces the degree to which the tests represent the running system, and can indicate a design that has been excessively broken up solely for the purposes of being able to inject mocks into various points of the operation.
 
+* Do not rely on test IDs of any kind. See the [test IDs](#test-ids) section below for alternative to test IDs, and read additional rationale for this decision in its [decision record](../Decision%20records/04%20-%20We%20do%20not%20use%20test%20IDs%20in%20tests.md).
+
 ### Setup, perform, assert
 
 All tests should follow a common pattern. First, do any necessary setup work (for example, instantiating an object, or mocking a dependency to return a particular value). Then, perform the action you are testing (for example, triggering a method). Finally, assert on the result, either on the value received from performing the action, or on the side effects that are expected as part of it.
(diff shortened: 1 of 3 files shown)


Best practices/Testing.md
```diff
@@ -205,12 +207,105 @@ Common patterns for organizing test and fixture files makes it easier to navigat
...
+
+* Similarly to the above, if you wish to assert that a property of a component is "injected" into the component’s end markup, assert that it appears *somewhere* in the component, without nailing down any particular element:
+
+  ```ts
+  // Example using React and Enzyme
+
+  it('includes the children in the markup', () => {
+    const children = <div>Contents</div>;
+    const card = mount(<Card>{children}</Card>);
+    expect(card).toContainReact(children);
```
tzvipm: TIL ❤️

Best practices/Testing.md
```diff
@@ -205,12 +207,105 @@ Common patterns for organizing test and fixture files makes it easier to navigat
...
+
+    // and, in your test, you can now find based on this property,
+    // shown here using Enzyme:
+
+    expect(myComponent.find(TextField).filter({id: 'NameField'})).toHaveProp('value', '');
+    ```
+
+    To make this approach even easier, our [`enzyme-utilities` package](https://github.com/Shopify/quilt/tree/master/packages/enzyme-utilities) provides a `findById` utility function.
+  
+  * If neither of the above apply, filter the set of matched components based on the properties that are unique to them. If no properties are unique, fall back to the index of the item in the matched set.
```
tzvipm: this wording is a bit confusing to me. To confirm, if you can't make it into a separate component, and a unique ID doesn't make sense, then simply find all instances of the component and filter them by some other unique thing like text or a prop. If there are no unique attributes that make sense, then locate the item by its index (ex. the second modal)?

Decision records/04 - We do not use test IDs in tests.md
```diff
@@ -0,0 +1,49 @@
+# We do not use test IDs in tests
```
tzvipm: maybe just `We do not use test IDs` ?

loic-d:
We ran into this problem with our `DiscountSummary` component that renders different summary lines to describe discount characteristics to merchants. However, based on the type of discount, the outputted markup for a given line will be different. We used `testID` to clearly identify which part of the JSX was rendered.

I think that removing `testID` would have forced us to maybe split our component in sub-components per summary lines, or just assert that the text at a given line index was correct. At the end of the day, we just want to know if the UI matches the UX spec, but don't really care how we got there.

The proposed changes make sense 👍 but we will definitely have to do some refactoring in multiple places to respect this new best practice.

Best practices/Testing.md
loic-d: Really nice, didn't know you could do that.

rox163:
Love it! Simple yet in-depth explanation of a really important practice.

Best practices/Testing.md
```diff
@@ -205,12 +207,105 @@ Common patterns for organizing test and fixture files makes it easier to navigat
...
+
+* `div`s and `span`s that have no semantic value, but contain some content (text, child components, etc) that *are* a guarantee the developer wants to make.
+* Markup that has an important effect on the visual output of a component, but which have no other defining characteristics (for example, an element on which we apply a particular class name or inline styles).
+* A child component that is rendered multiple times, where the developer wishes to easily disambiguate between the instances (for example, a component that renders multiple `Modal`s from Polaris React).
+
+We would like to avoid test IDs for handling the cases above, as make it easy to reach into your component in ways that unnecessarily lock down the implementation. You can typically avoid test IDs by using one of the following strategies:
```
cartogram: `as make it easy` => `as they make it easy`

Decision records/04 - We do not use test IDs in tests.md
```diff
@@ -0,0 +1,49 @@
...
+* Chris Sauvé
+* Mallory Allen
+
+## Summary
+
+Test IDs make it too easy to test implementation details of components. They should be avoided in favor of [alternatives](../Best%20practices/Testing.md#test-ids) that emphasize the guarantees a component actually makes.
+
+## Description
+
+Developers sometimes feel the need to add "hooks" that allow their tests to easily find parts of the component subtree and assert details about those subtrees in tests. In the React ecosystem, this is a key recommendation of an alternative to Enzyme (our preferred React testing library), [`react-testing-library`/ `dom-testing-library`](https://github.com/kentcdodds/dom-testing-library/blob/master/README.md#faq).
```
cartogram: That second sentence is distracting and does not really add anything.

Decision records/04 - We do not use test IDs in tests.md
```diff
@@ -0,0 +1,49 @@
...
+
+## History
+
+In building the initial version of Shopify Web, we took several steps to encourage and facilitate the use of test IDs:
+
+* A [babel plugin](https://github.com/lemonmade/babel-plugin-react-test-id) to remove the unnecessary attribute in production
+* Allowing the `testID` prop implicitly with TypeScript and adding a `findByTestId` utility to Shopify Web
```
cartogram: This is 2 separate bullets, but its fine.

Decision records/04 - We do not use test IDs in tests.md
```diff
@@ -0,0 +1,49 @@
...
+
+These were originally introduced to facilitate differentiation between the same React component rendered multiple times within a tree (specifically, to target multiple `TextField` components rendered by a card in the product section of the admin). No analysis of alternatives was performed at the time. The cases in which this attribute was used expanded and, as of July 10, 2018, it is used 280 times in the Web codebase.
```
cartogram: I don't think you need the part in parens. And maybe `Web` should be `Shopify Web` to match when it is mentioned above.

Decision records/04 - We do not use test IDs in tests.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Decision
+
+`testID` provides some benefits that have made it attractive to developers working in the Web stack:
```
cartogram: `Web` => `Shopify Web`

Decision records/04 - We do not use test IDs in tests.md
```diff
@@ -0,0 +1,49 @@
...
+* Our implementation in Shopify Web allows it to be used on any HTML element or custom component (without changes to the component’s API)
+* It has no runtime impact because the attribute is stripped in production builds
+
+However, implementing `testID` for a project involves several steps, and all of them are clearly monkey patches:
+
+* You must augment an internal React interface for TypeScript to not complain about the unexpected attribute
+* You must prevent console warnings that React prints by default complaining about an unexpected `testID` prop on DOM elements
+* You must be sure to include our custom Babel plugin to strip the attribute in production builds
+
+Additionally, in reviewing the use of `testID` in Web, we found that the ways it were used were better addressed using other strategies. We have detailed common patterns we found (along with alternative strategies to `testID`) in our [testing best practices](../Best%20practices/Testing.md#test-ids), but in summary, we found that `testID` was used to:
```
cartogram: `Web` => `Shopify Web` (or change them all to just `Web`). I don't have a preference other than that it is consistent.

Best practices/Testing.md
lemonmade: Yes, that is exactly correct (text is always a prop of the component, but I wanted to try to not tie this entire section just to React)

PR #66, opened 2018-11-20
Add a few testing guidelines

solonaarmstrong:
Closes: https://github.com/Shopify/web-foundation/issues/64

There are a few preferences that keep coming up in Polaris and I would like to explicitly add them to the style guide.

### DIFF
--- a/Best practices/Enzyme.md
+++ b/Best practices/Enzyme.md
@@ -78,3 +78,27 @@ expect(myComponent.find(ChildComponent)).toHaveLength(1);
 // over:
 expect(myComponent.find(ChildComponent).exists()).toBe(true);
 ```
+
+## Event simulations
+
+* When triggering events on a composite component, use `trigger` on the prop; when doing it for raw DOM elements, use `simulate`.
+
+```js
+const myComponent = mount(<MyComponent />);
+
+// prefer:
+trigger(myComponent.find(SomeComponent), 'onClick');
+
+// over:
+myComponent.find('button').simulate('click');
+```
+
+```js
+const myComponent = mount(<MyComponent />);
+
+// prefer:
+myComponent.find('button').simulate('click');
+
+// over:
+trigger(myComponent.find(SomeComponent), 'onClick');
+```
--- a/Best practices/React/Testing.md
+++ b/Best practices/React/Testing.md
@@ -98,6 +98,26 @@ Another benefit of testing React components is that they typically abstract away
   describe('<MyComponent />', () => {});
   ```
 
+* When there are multiple tests for the same prop, they should be wrapped in a single describe block.
+
+  ```js
+  // bad
+  describe('<MyComponent />', () => {
+    it('disabled gets passed to <MySubComponent />', () => {});
+    it('renders if disabled is true', () => {});
+    it('does not render if disabled is false', () => {});
+  });
+
+  // good
+  describe('<MyComponent />', () => {
+    describe('disabled', () => {
+      it('gets passed to <MySubComponent />', () => {});
+      it('renders if true', () => {});
+      it('does not render if false', () => {});
+    });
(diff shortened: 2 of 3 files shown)


Best practices/Enzyme.md
```diff
@@ -8,6 +8,28 @@ We have one utility package for Enzyme, [@shopify/enzyme-utilities](https://gith
...
+  // bad
+  describe('<MyComponent />', () => {
+    it('MyComponent passes the id prop to ChildComponent', () => {});
+  });
+
+  // good
+  describe('<MyComponent />', () => {
+    it('passes the id prop to ChildComponent', () => {});
+  });
+  ```
```
lemonmade: I think these are maybe more appropriate in the general testing guide?

Best practices/Enzyme.md
```diff
@@ -78,3 +100,21 @@ expect(myComponent.find(ChildComponent)).toHaveLength(1);
 // over:
 expect(myComponent.find(ChildComponent).exists()).toBe(true);
 ```
+
+## Event simulations
+
+* When triggering events, trigger the callback function for the event, whenever possible.
```
lemonmade: I think you can combine these into one rule; when triggering props on a composite component, use `trigger`; when doing it for native DOM events, use `simulate`

Best practices/React/Testing.md
```diff
@@ -98,6 +98,18 @@ Another benefit of testing React components is that they typically abstract away
   describe('<MyComponent />', () => {});
   ```
 
+* When there are multiple tests for the same prop or callback function, they should be wrapped in a single describe block
```
lemonmade: Needs punctuation

lemonmade: Also, callback functions are just props, so I might just leave that out.

Best practices/React/Testing.md
```diff
@@ -98,6 +98,18 @@ Another benefit of testing React components is that they typically abstract away
...
+
+  ```js
+  describe('<MyComponent />', () => {
```
lemonmade: Should we should a "bad" example here?

lemonmade: This one is also kind of referenced in https://github.com/Shopify/web-foundation/pull/66/files#diff-eeb8827fd304d15a55965bd445d003e0R9, should we just add the example there?

solonaarmstrong: I think it's a good idea to call this one out explicitly with an example.

Best practices/Enzyme.md
lemonmade: You don't need to construct a valid mock event object

Best practices/Enzyme.md
```diff
@@ -78,3 +78,19 @@ expect(myComponent.find(ChildComponent)).toHaveLength(1);
...
+const myComponent = mount(<MyComponent />);
+
+trigger(myComponent, 'onClick');
+```
+
+```js
+const myComponent = mount(<MyComponent />);
+
+myComponent.simulate('click');
+```
```
lemonmade: Needs a newline, might be worth having the counter examples for each of these too.

Best practices/React/Testing.md
```diff
@@ -98,6 +98,48 @@ Another benefit of testing React components is that they typically abstract away
...
+    describe('disabled', () => {
+      it('gets passed to <MySubComponent />', () => {});
+      it('renders if true', () => {});
+      it('does not render if false', () => {});
+    });
+  });
+  ```
+
+* The statement in an it block should use present tense without repeating what is being described.
```
lemonmade: I think this should go in the main testing guide (sorry if I am waffling)

solonaarmstrong: No worries.

PR #77, opened 2019-04-08
Add basic hook docs

lemonmade:
Fixes https://github.com/Shopify/web-foundation/issues/75
Fixes https://github.com/Shopify/web-foundation/issues/74
Fixes https://github.com/Shopify/web-foundation/issues/76

This PR adds a bunch of docs to support hook stuff:

1. Update testing guide to point at new hook-friendly libraries
1. Add API design doc that gives some design guidelines for hooks
1. Document the shape of an `app/hooks` directory for basic hook composition
1. Document the general shape of a library that has hook + context + other stuff
1. Document preference for testing components using the hook rather than the hook directly

The only one I am still waffling on is the `app/hooks`/ `MyComponent/hooks` thing. Conceptually I like the hooks being there, because hooks can be equally important to components (which we have `app/components`/ `MyComponent/components` for), but I worry that there will be a lot of reaching between `utilities` and `hooks` directories.

### DIFF
--- a/Best practices/React/Testing.md
+++ b/Best practices/React/Testing.md
@@ -89,6 +89,8 @@ There are many parts of React components that should not be tested as they are i
 
 * `state`. State is not directly manipulable from other components, and so should be treated as private. In order to simulate setting state, trigger the props of components rendered by your component with arguments that result in the component getting into that state. Once a component is in a state, do not assert on the state having a particular shape; instead, assert that the subcomponents being rendered have the expected props.
 
+* What hooks were used in a component. The hooks you use generally end up doing one of three things: pull in context, store local state, or create side effects. How context is pulled in to a component is an implementation detail; `useContext`, `contextType`, and `Context.Consumer` are all acceptable and interchangeable. As noted above, local state is never tested. Finally, we should test that the side effects themselves happened, not how they were triggered.
+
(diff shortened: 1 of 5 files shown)


Best practices/React/Testing.md
```diff
@@ -84,7 +89,7 @@ There are many parts of React components that should not be tested as they are i
 
 * `state`. State is not directly manipulable from other components, and so should be treated as private. In order to simulate setting state, trigger the props of components rendered by your component with arguments that result in the component getting into that state. Once a component is in a state, do not assert on the state having a particular shape; instead, assert that the subcomponents being rendered have the expected props.
 
-* Instances for class components. All methods should be private on your React components, as you will otherwise be encouraging users to break out of the declarative model of React. This includes React’s lifecycle methods, which are implementation details of the framework itself. Never call any of these methods directly.
+* Instances for class components (or of function components, through `React.useImperativeHandle`). All methods should be private on your React components, as you will otherwise be encouraging users to break out of the declarative model of React. This includes React’s lifecycle methods, which are implementation details of the framework itself. Never call any of these methods directly.
```
TheMallen: I feel like this is a bit mixed in it's messaging since `useImperativeHandle` is explicitly for defining these types of methods, whereas class methods appearing on the instance is a natural result of programming using classes.

I'm not sure there's ever a reason to call `useImperativeHandle` and pass in an object with private methods, so I'd prefer a blanket recommendation against using it.

PR #83, opened 2019-07-19
🗒️ Add Decision Record for React SSR

TheMallen:
Not meant to go in until we actually have all of this fully in place, just wanted to have a place to link folks and get the conversation started on this document.


Decision records/10 - We use dedicated node services for server-side-rendering React in Rails applications.md
```diff
@@ -0,0 +1,37 @@
...
+
+Our solution is actually quite similar to Airbnb's [`HyperNova`](https://github.com/airbnb/hypernova), however that library has it's own opinionated runtime associated which precludes the use of arbitrary serialization, and instead focuses on serializing state for specific libraries, as well as server-provided props for the top-level component. In addition, Airbnb itself has moved away from [`HyperNova`](https://github.com/airbnb/hypernova) citing concerns with having to write process supervisor trees in application code for larger projects. Since our solution is built from the ground up around letting the application code define it's own serialization / deserialization needs, and make use of our service scaling infrastructure, we can avoid the problems they encountered.
```
kaelig:
```suggestion
Our solution is actually quite similar to Airbnb's [`HyperNova`](https://github.com/airbnb/hypernova), however that library has its own opinionated runtime associated which precludes the use of arbitrary serialization, and instead focuses on serializing state for specific libraries, as well as server-provided props for the top-level component. In addition, Airbnb itself has moved away from [`HyperNova`](https://github.com/airbnb/hypernova) citing concerns with having to write process supervisor trees in application code for larger projects. Since our solution is built from the ground up around letting the application code define its own serialization / deserialization needs, and make use of our service scaling infrastructure, we can avoid the problems they encountered.
```

PR #145, opened 2020-03-31
Remove react-hot-loader

jtsmills:
As part of https://github.com/Shopify/sewing-kit/issues/1777 and the implementation of https://github.com/Shopify/sewing-kit/pull/1952 we wish to update to `react-fast-refresh` as provided by https://github.com/pmmmwh/react-refresh-webpack-plugin. This requires removing `react-hot-loader`.

This requires the usage of a new babel-transform `react-refresh/babel` and the aforementioned `react-refresh-webpack-plugin`. We are choosing to perform the logic to include these two packages within sewing-kit because of their dependent nature.

### DIFF
--- a/packages/babel-preset/README.md
+++ b/packages/babel-preset/README.md
@@ -95,7 +95,6 @@ This packages comes with several different presets for you to use, depending on 
 - `@shopify/babel-preset/react`: Adds plugins that transform React (including JSX). You can use this preset with the `@shopify/babel-preset/web` or `@shopify/babel-preset/node` configuration.
 
   This preset accepts an options object.
-  - `hot` : Will automatically add plugins to enable hot reloading of React components. Note that this requires you to have a recent version of `react-hot-loader` installed as a dependency in your project.
   - `pragma` : Replace the function used when compiling JSX expressions. Defaults to `React.createElement`.
   - `pragmaFrag`: Replace the function used when compiling JSX fragment expressions. Defaults to `React.Fragment`.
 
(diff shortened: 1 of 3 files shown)


BPScott:
Don't forget to update the README to remove mention of the hot option

packages/babel-preset/README.md
```diff
@@ -95,15 +95,14 @@ This packages comes with several different presets for you to use, depending on
...
-  - `hot` : Will automatically add plugins to enable hot reloading of React components. Note that this requires you to have a recent version of `react-hot-loader` installed as a dependency in your project.
   - `pragma` : Replace the function used when compiling JSX expressions. Defaults to `React.createElement`.
   - `pragmaFrag`: Replace the function used when compiling JSX fragment expressions. Defaults to `React.Fragment`.
 
   ```json
   {
     "babel": {
       "presets": [
-        ["@shopify/babel-preset/react", {"hot": true}]
+        ["@shopify/babel-preset/react", {"pragma": "React.createElement"}]
```
lemonmade, line 105: Why the change to the pragma?

jtsmills: Was leaving this as an example of changing the options to the preset. Since hot is no longer an option

BPScott: Showcasing "here's how you change the options from the defaults" by specifying the default value for the option  feels a bit odd to me.

If you feel strongly about having an example here perhaps `{"pragma": "h"}` for preact usage?

BPScott:
Oh, slightly late - when you release this, don't forget to add an entry to the changelog: https://github.com/Shopify/web-foundation/blob/master/packages/babel-preset/CHANGELOG.md as a major change

PR #150, opened 2020-04-30
[eslint-plugin] adding new options to strict-component-boundaries

patsissons:
adding the following options to `strict-component-boundaries` list rule
- `allow` (`string[]`): list of patterns to automatically allow (i.e., `['section/Foo/.*']`)
- `maxDepth` (`number`): max allowed inclusive depth for `components` boundary (default is `1`)
  - `components` would be depth `1` since there is a single inclusive `components` level 
  - `components/Foo` would be depth `2`
  - `sections/Foo/components/Bar/Baz` would be depth `3`

### DIFF
--- a/packages/eslint-plugin/lib/rules/strict-component-boundaries.js
+++ b/packages/eslint-plugin/lib/rules/strict-component-boundaries.js
@@ -14,8 +14,30 @@ module.exports = {
       uri: docsUrl('strict-component-boundaries'),
     },
     fixable: null,
+    schema: [
+      {
+        type: 'object',
+        properties: {
+          allow: {
+            type: 'array',
+            items: {type: 'string'},
+          },
+          maxDepth: {
+            type: 'integer',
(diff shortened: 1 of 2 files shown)


alexandcote:
Do you know if the regex can match the absolute path?

Let say we want to allow:
```
sections/Editor/components/Foo
```

Should we also allow : 
```
../components/Foo
```

Maybe you already handle this? 🤔

patsissons:
> Do you know if the regex can match the absolute path?
> 
> Let say we want to allow:
> 
> ```
> sections/Editor/components/Foo
> ```
> 
> Should we also allow :
> 
> ```
> ../components/Foo
> ```
> 
> Maybe you already handle this? 🤔

we should be able to handle this via

```
options: [{allow: [
  // relative support
  'components/\\w+$',

  // absolute support
  '^sections/.*/components/\\w+$',
]}],
```

PR #152, opened 2020-05-08
Create shared prettier config package

janiceblundell:
Fixes [issue# 147](https://github.com/Shopify/web-foundation/issues/147)
Moving existing prettier config to a shared web-foundation package that can be referenced, eliminating the need for an individual project `.prettierrc`.
 
Reference documentation for shared prettier config:
https://prettier.io/docs/en/configuration.html#sharing-configurations

This shared configuration integrates with existing ESLint and stylelint configs
https://www.npmjs.com/package/@shopify/eslint-plugin
https://www.npmjs.com/package/@shopify/stylelint-plugin

### DIFF
--- a/packages/prettier-config/README.md
+++ b/packages/prettier-config/README.md
@@ -1,0 +1,14 @@
+# `@shopify/prettier-config`
+
+[![Build Status](https://travis-ci.com/Shopify/web-foundation.svg?branch=master)](https://travis-ci.com/Shopify/web-foundation)
+[![License: MIT](https://img.shields.io/badge/License-MIT-green.svg)](../../LICENSE.md) [![npm version](https://badge.fury.io/js/%40shopify%2Fprettier-config.svg)](https://badge.fury.io/js/%40shopify%2Fprettier-config.svg)  [![npm bundle size (minified + gzip)](https://img.shields.io/bundlephobia/minzip/@shopify/prettier-config.svg)](https://img.shields.io/bundlephobia/minzip/@shopify/prettier-config.svg) 
+
+Shared prettier configuration
+
+## Installation
+
+```bash
+$ yarn add @shopify/prettier-config
+```
+
+## Usage
--- a/packages/prettier-config/README.md
+++ b/packages/prettier-config/README.md
@@ -8,7 +8,14 @@ Shared prettier configuration
 ## Installation
 
 ```bash
-$ yarn add @shopify/prettier-config
+$ yarn add --dev @shopify/prettier-config
 ```
 
 ## Usage
+Shopify’s shared prettier config comes bundled in `@shopify/prettier-config`. To enable these rules, add a `prettier` property in your `package.json` and reference this shared config as follows:
+```
+"prettier": "@shopify/prettier-config"
+```
+Previously, rules had been defined directly in a `.prettierrc` or `package.json` 
+
+Any previous `.prettierrc` should be removed in favour of the shared config.
--- a/packages/prettier-config/index.json
+++ b/packages/prettier-config/index.json
@@ -1,0 +1,6 @@
+{
+  "arrowParens": "always",
+  "singleQuote": true,
+  "bracketSpacing": false,
+  "trailingComma": "all"
+}
--- a/packages/prettier-config/index.json
+++ b/packages/prettier-config/index.json
@@ -2,5 +2,6 @@
   "arrowParens": "always",
   "singleQuote": true,
   "bracketSpacing": false,
-  "trailingComma": "all"
+  "trailingComma": "all",
+  "quoteProps": "as-needed"
 }
--- a/packages/prettier-config/index.json
+++ b/packages/prettier-config/index.json
@@ -1,7 +1,5 @@
 {
-  "arrowParens": "always",
   "singleQuote": true,
   "bracketSpacing": false,
-  "trailingComma": "all",
-  "quoteProps": "as-needed"
+  "trailingComma": "all"
 }
--- a/yarn.lock
+++ b/yarn.lock
@@ -1808,11 +1808,6 @@ bcrypt-pbkdf@^1.0.0:
   dependencies:
     tweetnacl "^0.14.3"
 
-big.js@^5.2.2:
-  version "5.2.2"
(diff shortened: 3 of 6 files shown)


yarn.lock
```diff
@@ -1808,11 +1808,6 @@ bcrypt-pbkdf@^1.0.0:
   dependencies:
     tweetnacl "^0.14.3"
 
-big.js@^5.2.2:
```
ismail-syed, line 1811: Why are there `yarn.lock` changes? We're not changing any dependencies 😕

packages/prettier-config/index.json
```diff
@@ -0,0 +1,6 @@
+{
```
ismail-syed, line 1: What about the [`quoteProps`](https://github.com/Shopify/quilt/blob/master/.prettierrc#L6) rule?

janiceblundell: Added.

yarn.lock
janiceblundell: I actually don't know.  I just created a completely new branch off master to start over, and when I dev up the yarn.lock is changed.

ismail-syed: Ahh, interesting. I was able to re-produce with a clean master with a `dev up`. Makes sense 👍

packages/prettier-config/README.md
```diff
@@ -0,0 +1,14 @@
...
+
+Shared prettier configuration
+
+## Installation
+
+```bash
+$ yarn add @shopify/prettier-config
+```
+
+## Usage
```
ismail-syed, line 14: Can we fill this out, please? It might be worth adding notes on how this package is intended to be used with `@shopify/eslint-plugin`. Most of it is "automatic" but it's still worth documenting the automation.

packages/prettier-config/README.md
```diff
@@ -0,0 +1,14 @@
...
+
+[![Build Status](https://travis-ci.com/Shopify/web-foundation.svg?branch=master)](https://travis-ci.com/Shopify/web-foundation)
+[![License: MIT](https://img.shields.io/badge/License-MIT-green.svg)](../../LICENSE.md) [![npm version](https://badge.fury.io/js/%40shopify%2Fprettier-config.svg)](https://badge.fury.io/js/%40shopify%2Fprettier-config.svg)  [![npm bundle size (minified + gzip)](https://img.shields.io/bundlephobia/minzip/@shopify/prettier-config.svg)](https://img.shields.io/bundlephobia/minzip/@shopify/prettier-config.svg) 
+
+Shared prettier configuration
+
+## Installation
+
+```bash
+$ yarn add @shopify/prettier-config
```
janiceblundell, line 11: Should this actually say `yarn add --dev @shopify/prettier-config` as the documentation for `@shopify/eslint-plugin` and `@shopify/stylelint-plugin` does?

ismail-syed: Yes, and also mention that it can be used by adding a `prettier` field in a project's `package.json`.

packages/prettier-config/index.json
BPScott: `"quoteProps": "as-needed"` and `"arrowParens": "always"` are prettier defaults (as of v2 at least). Do we need to specify those if they're already the default?s

janiceblundell: No, I guess those can be removed in that case.  Will test.

packages/prettier-config/README.md
janiceblundell: Done.

packages/prettier-config/README.md
```diff
@@ -0,0 +1,76 @@
...
+```
+
+## Usage
+Shopify’s shared prettier config comes bundled in `@shopify/prettier-config`. To enable these rules, add a `prettier` property in your `package.json` and reference this shared config as follows:
+```
+"prettier": "@shopify/prettier-config"
+```
+Previously, rules had been defined directly in the `package.json` 
+```
+"prettier": {
```
ismail-syed, line 21: I don't think we'll need to list out these snippets. 

```
Previously, rules have been explicitly defined directly in a `.prettierrc` or `package.json` 
```

packages/prettier-config/README.md
```diff
@@ -0,0 +1,76 @@
...
+```
+{
+  "arrowParens": "avoid",
+  "singleQuote": true,
+  "bracketSpacing": false,
+  "trailingComma": "all",
+  "quoteProps": "as-needed"
+}
+```
+Any previous .prettierrc should be removed in favour of the shared config.
```
ismail-syed, line 39: `.prettierrc`

packages/prettier-config/README.md
```diff
@@ -0,0 +1,76 @@
...
+See [configuring prettier](https://prettier.io/docs/en/configuration.html) and [sharing configuration](https://prettier.io/docs/en/configuration.html#sharing-configurations) for more details.  
+
+This shared prettier config integrates with [@shopify/eslint-plugin](https://github.com/Shopify/web-foundation/blob/master/packages/eslint-plugin/README.md) and [@shopify/stylelint-plugin](https://github.com/Shopify/web-foundation/blob/master/packages/stylelint-plugin/README.md).
+
+The following is an example configuration using `eslint-plugin`, `stylelint-plugin` and shared `prettier-config`:
```
ismail-syed, line 44: I think we can omit this section and code snippet. If folks are interested, they can explore the code. Large snippets like this can get stale.

packages/prettier-config/README.md
```diff
@@ -0,0 +1,76 @@
...
+  "singleQuote": true,
+  "bracketSpacing": false,
+  "trailingComma": "all",
+  "quoteProps": "as-needed"
+}
+```
+Any previous .prettierrc should be removed in favour of the shared config.
+See [configuring prettier](https://prettier.io/docs/en/configuration.html) and [sharing configuration](https://prettier.io/docs/en/configuration.html#sharing-configurations) for more details.  
+
+This shared prettier config integrates with [@shopify/eslint-plugin](https://github.com/Shopify/web-foundation/blob/master/packages/eslint-plugin/README.md) and [@shopify/stylelint-plugin](https://github.com/Shopify/web-foundation/blob/master/packages/stylelint-plugin/README.md).
```
ismail-syed, line 42: Can we add this in after we integrate with the tools instead? It may cause some confusion until then.

PR #158, opened 2020-05-19
Allow JSX on files with .jsx extension

volmer:
Correct me if I'm wrong but the idea here would be to allow JSX in `.jsx` files, no?


lemonmade:
Our [styleguide](https://github.com/Shopify/javascript/tree/master/react#1.1) recommends using .js files, even when JSX is included. This seems to be the more common approach in the community, including its usage as the default for projects like create-react-app and Next.

volmer:
Thanks for clarifying! I assumed it was just a config typo.

PR #160, opened 2020-05-26
Add quilt decision records to web-foundation

ismail-syed:
Quilt's decision records are being moved here to consolidated our org-wide decision records.


alexandcote (approved):
Can you remove the decision log from quilt and maybe add a file that points here instead?

ismail-syed:
yes, already being done here, https://github.com/Shopify/quilt/pull/1462

PR #165, opened 2020-06-04
Update scss type

vsumner:
Update the scss import type.


vsumner:
> I for one welcome our new default export using overlords.
> 
> Add a changelog entry and we're good to go.
> 
> I'm thinking this is a major version bump as we wouldn't want folks on older SK versions - that don't have css-loader configured with `esModules:true` - to pull this change in

Agreed.

PR #168, opened 2020-06-25
[ESLint-plugin] Adds 'getDerivedStateFromError' to the list of React lifecycle methods to 'react-prefer-private-members'

antoinegrant:
Adds `getDerivedStateFromError` to the list of React lifecycle methods to `@shopify/react-prefer-private-members`.

Fixes https://github.com/Shopify/web-foundation/issues/167

To allow:
```ts
class ErrorBoundary extends Component<PropsWithChildren<Props>, State> {
  static getDerivedStateFromError() {
    return {hasError: true};
  }

  render() {
    ...
  }
}
```


BPScott:
Merge away! Ask in #web-foundation-tech if you want a hand releasing this

PR #171, opened 2020-07-09
Fix broken stuff caused by default branch rename

movermeyer:
The default branch was changed from `master` to `main`, which broke all the links, badges, and also the `docsUrl` function.


movermeyer:
@lemonmade 

Can I leave this to your team to merge? Given that I don't know what your deploy process is like.

tleunen:
There's no need to deploy anything imo. But in any case. Pinging the team to let them know of these changes ;) @Shopify/web-foundation

PR #172, opened 2020-07-13
Update babel to v7.10.4, and enable bugfixes mode

BPScott:
See https://babeljs.io/docs/en/babel-preset-env#bugfixes

7.10 also adds support for better tree-shaking of react components: https://babeljs.io/blog/2020/05/25/7.10.0


packages/babel-preset/non-standard-plugins.js
```diff
@@ -16,14 +16,19 @@ module.exports = function shopifyNonStandardPlugins(options = {}) {
     // proposal-class-properties must be used in loose mode
     // see https://babeljs.io/docs/en/babel-plugin-proposal-decorators#note-compatibility-with-babel-plugin-proposal-class-properties
     plugins.push(
-      require.resolve('@babel/plugin-syntax-nullish-coalescing-operator'),
```
BPScott, line 19: syntax support for nullish coalescing / optional chaining are already added by babel-preset-env, so no need to reference them here or in our package.json

PR #175, opened 2020-08-17
Bump @typescript-eslint to support TS4

jtsmills:
Bump https://github.com/typescript-eslint/typescript-eslint to `3.9.1` to land support for TS4. There are some [some breaking changes with the move from 2 to 3.](https://github.com/typescript-eslint/typescript-eslint/releases/tag/v3.0.0)

Currently using a [resolution on web](https://github.com/Shopify/web/pull/30255).

Updating to using the `naming-convention` rule also required some changes. According to [this @typescript-eslint issue](https://github.com/typescript-eslint/typescript-eslint/issues/2185), the old `camelCase` rule simply checked if there was an underscore internal to the variable name. As such, our styling has allowed PascalCase in var names etc (also important for React classes). I am enabling PascalCase for `variable` and we can revisit this in the future.

expected allowed:

```
  const myVariable = 3;
  const COOL_CONST = "hey";
  export function ReactComponent() { ... }
  const [Toast, showToast] = useToast();
```

expected disallowed:

```
  const not_cool = "darn"
```

When testing in web the output looked okay. Expected errors came from not being able to find the deprecated rules for `ignore` lines.

### DIFF
--- a/packages/eslint-plugin/CHANGELOG.md
+++ b/packages/eslint-plugin/CHANGELOG.md
@@ -7,6 +7,17 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 <!-- ## [Unreleased] -->
 
+## [Unreleased]
+
+### Breaking Change
+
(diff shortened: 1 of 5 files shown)


BPScott:
Can we remove that mention of `eslint-plugin-typescript` in our package.json? It looks like it's been [replaced by @typescript-eslint/eslint-plugin](https://www.npmjs.com/package/eslint-plugin-typescript)?

jtsmills:
Made a final pass after running it against web again.

Looks like we have some failures for new `positive failures` for variables shaped like `process.env.IS_STAGING`. But I think this is the setup to go forward with.

PR #176, opened 2020-08-24
Prefer buildClientSchema to buildSchema

keyfer:
Added a new rule, `@shopify/typescript/prefer-build-client-schema`, which flags uses of `buildSchema` and recommends using `buildClientSchema` instead due to our findings in performance.

For more information on the performance findings, check [here](https://github.com/Shopify/web/pull/30292).

### DIFF
--- a/packages/eslint-plugin/lib/rules/typescript/prefer-build-client-schema.js
+++ b/packages/eslint-plugin/lib/rules/typescript/prefer-build-client-schema.js
@@ -1,0 +1,33 @@
+const {docsUrl} = require('../../utilities');
+
+module.exports = {
+  meta: {
+    docs: {
+      description: 'Prefer buildClientSchema for schema building',
+      category: 'Best Practices',
+      recommended: true,
+      uri: docsUrl('typescript/prefer-build-client-schema'),
+    },
+    fixable: null,
+  },
+  create(context) {
+    function report(node) {
+      context.report({
+        node,
+        message: `Prefer buildClientSchema to buildSchema`,
+      });
+    }
+
+    return {
+      ImportDeclaration(node) {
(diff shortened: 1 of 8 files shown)


packages/eslint-plugin/lib/rules/typescript/prefer-build-client-schema.js
```diff
@@ -0,0 +1,33 @@
...
+      recommended: true,
+      uri: docsUrl('typescript/prefer-build-client-schema'),
+    },
+    fixable: null,
+  },
+  create(context) {
+    function report(node) {
+      context.report({
+        node,
+        message: `Prefer buildClientSchema to buildSchema`,
```
jtsmills, line 17: I wonder if we could even apply the fix here?

https://eslint.org/docs/developer-guide/working-with-rules#applying-fixes-1

jtsmills (approved):
⛵ Looks good to me! Thanks for grabbing this Keyfer!

packages/eslint-plugin/lib/rules/typescript/prefer-build-client-schema.js
keyfer: I was thinking about this, but due to not knowing how `buildSchema` could be used, and the differences in implementation, applying a standardized fix across `x` number of repos, with `x` implementations could get weird.

jmreid:
Looks like new Rails apps generated with `dev init` will fail `yarn lint` out of the box now:
https://github.com/Shopify/shopify-rails/blob/c764ea5ebbbc412d469c13db0541b14b31eda4b4/lib/shopify/rails/templates/shopify/web/tests/graphql.ts#L5

Issue #126, opened 2018-08-29
Jest rule for lines between it blocks


ismail-syed:
Add a jest rule for lines between it blocks in tests. There's a similar [rule](https://eslint.org/docs/rules/lines-between-class-members) one in eslint for class memebers. 

Something like:

```
//bad
describe('some feature', () => {
  it('does foo', () => {
    // foo things 
  });
  it('does bar', () => {
    // bar things
  });
});


//good
describe('some feature', () => {
  it('does foo', () => {
    // foo things 
  });

  it('does bar', () => {
    // bar things
  });
});
```

jgodson:
I tried out [this one](https://www.npmjs.com/package/eslint-plugin-jest-formatting) and it seems to be great. The strict preset enables all the rules, which all make sense to me. I'll PR it 🙂

ryanwilsonperkin:
Hey, my buddy made that package! I don't have strong feelings about this particular lint rule, but I think it's well-built

PR #178, opened 2020-09-09
New eslint default for jest formatting

ryanwilsonperkin:
Addresses #126

Introduces the [eslint-plugin-jest-formatting](https://www.npmjs.com/package/eslint-plugin-jest-formatting) eslint plugin and configures the recommended settings to ship as part of our `@shopify/jest` eslint plugin.

This enforces some rules about the code styling of the blocks used in jest tests, for example: `jest-formatting/padding-around-after-all-blocks` which states:

Examples of **incorrect** code for this rule:

```js
const someText = 'abc';
afterAll(() => {});
describe('someText', () => {});
```

Examples of **correct** code for this rule:

```js
const someText = 'abc';

afterAll(() => {});

describe('someText', () => {});
```

### DIFF
--- a/packages/eslint-plugin/lib/config/rules/jest-formatting.js
+++ b/packages/eslint-plugin/lib/config/rules/jest-formatting.js
@@ -1,0 +1,16 @@
+// see https://github.com/dangreenisrael/eslint-plugin-jest-formatting#usage
+
+module.exports = {
+  // Require padding around afterAll blocks
+  'jest-formatting/padding-around-after-all-blocks': 'error',
+  // Require padding around afterEach blocks
+  'jest-formatting/padding-around-after-each-blocks': 'error',
+  // Require padding around beforeAll blocks
+  'jest-formatting/padding-around-before-all-blocks': 'error',
+  // Require padding around beforeEach blocks
+  'jest-formatting/padding-around-before-each-blocks': 'error',
+  // Require padding around describe blocks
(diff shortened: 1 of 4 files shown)


packages/eslint-plugin/lib/config/rules/jest-formatting.js
```diff
@@ -0,0 +1,16 @@
+// see https://github.com/dangreenisrael/eslint-plugin-jest-formatting#usage
+
+module.exports = {
+  // Require padding around afterAll blocks
```
jgodson, line 4: I like this one too https://github.com/dangreenisrael/eslint-plugin-jest-formatting/blob/HEAD/docs/rules/padding-around-expect-groups.md#padding-around-expect-groups

jgodson (approved):
Thanks for getting this up. I completely forgot.

packages/eslint-plugin/lib/config/rules/jest-formatting.js
ryanwilsonperkin: Yeah I think that one comes as part of the `strict` config, and I only included those from the `recommended` config, so as to be less contentious. We can add it in too, but I suspect that'll be the one that people might be more likely to disable based on personal preference

ryanwilsonperkin:
Thanks for the review @jgodson, this is my first time contributing to this repo, anything I should do aside from just merging?

keyfer (approved):
Looks good! Thanks for contributing 🎉

PR #183, opened 2020-09-11
Use Stale bot instead of a custom action

atesgoral:
## Description

Use Stale like we do in sewing-kit. The current action only tagged issues as stale and never closed them. Will remove the existing labels and let Stale handle the rest.


.github/stale.yml
```diff
@@ -0,0 +1,17 @@
...
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - pinned
+  - security
+# Label to use when marking an issue as stale
+staleLabel: stale
+# Comment to post when marking an issue as stale. Set to `false` to disable
+markComment: >
+  This issue has been automatically marked as stale because it has not had
+  recent activity. It will be closed if no further activity occurs. Thank you
```
gmcgibbon, line 14:
```suggestion
  recent activity. It will be closed in 7 days if no further activity occurs. Thank you
```

gmcgibbon (approved):

PR #184, opened 2020-09-18
🥼🧪✨ Part 1 - Change to Web-Configs

keyfer:
## Description

This PR aims to begin the process of rename, restructuring and rebuilding this repo to its mandate of providing internal and external developers with an opinionated set of common configuration to build web UI. 

### Updates

- Renaming packages and references to `@shopify/web-foundations` to `@shopify/web-configs`

    - _To make sure we don't break anything, we'll make the repository name change first, and then ship this PR immediately after._

- Change all references from `Web Foundation` to `Web Foundations`, which is our team's name

- Removed some references to Shopify based, internal-only tools, as an open-sourced repository should aim for internal/external parity and open collaboration when possible

- Remove broken and unused TravisCI build badges

### Next Steps

With this being a public-first repository, our goal should be to remove our internal-only tools and documents.

- Move away from Shipit for releasing
- Remove/Reorganize our development handbook, and team handbook, to better align with the mission of this repository
- Update essential package documentation

## Checklist

- [x] I have added a changelog entry, prefixed by the type of change noted above (Documentation fix and Test update does not need a changelog as we do not publish new version)

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -1,29 +1,27 @@
-# Web Foundation: a contributor's guide
+# Web Configs: a contributor's guide
 
 This guide is tailored to Shopifolk, although we welcome contributions from [the broader development community](#external-contributors) as well.
 
 ## [Code of conduct](./CODE_OF_CONDUCT.md).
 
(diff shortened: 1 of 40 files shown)


vsumner (approved):
Looks good to me!

sanjanadesai27 (approved):
lgtm but I don't have the strongest opinions on naming stuff

gmcgibbon (approved):
👍 web-configs is a better name for this repo

PR #186, opened 2020-09-22
Remove tophat

keyfer:
## Description

Since this repo is OSS, we should remove our tophat references since they're not available to our community.


package.json
```diff
@@ -14,11 +14,9 @@
     "generate": "plop",
     "release": "yarn lint:changelogs && lerna publish",
     "release-beta": "lerna publish --dist-tag beta",
-    "generate:package": "plop package && yarn plop docs",
-    "tophat": "tophat"
```
gmcgibbon, line 18: I don't understand, why is this not available? https://www.npmjs.com/package/@shopify/tophat is available on npm with a readme.

Issue #187, opened 2020-09-30
Stylelint deprecated rules


tleunen:
Currently having these warnings when running a project with the default styling preset

```
Deprecation Warning: 'at-rule-blacklist' has been deprecated. Instead use 'at-rule-disallowed-list'. See: https://github.com/stylelint/stylelint/blob/13.7.0/lib/rules/at-rule-blacklist/README.md
Deprecation Warning: 'declaration-property-unit-blacklist' has been deprecated. Instead use 'declaration-property-unit-disallowed-list'. See: https://github.com/stylelint/stylelint/blob/13.7.0/lib/rules/declaration-property-unit-blacklist/README.md
Deprecation Warning: 'declaration-property-value-blacklist' has been deprecated. Instead use 'declaration-property-value-disallowed-list'. See: https://github.com/stylelint/stylelint/blob/13.7.0/lib/rules/declaration-property-value-blacklist/README.md
Deprecation Warning: 'declaration-property-value-whitelist' has been deprecated. Instead use 'declaration-property-value-allowed-list'. See: https://github.com/stylelint/stylelint/blob/13.7.0/lib/rules/declaration-property-value-whitelist/README.md
Deprecation Warning: 'function-url-scheme-whitelist' has been deprecated. Instead use 'function-url-scheme-allowed-list'. See: https://github.com/stylelint/stylelint/blob/13.7.0/lib/rules/function-url-scheme-whitelist/README.md
Deprecation Warning: 'property-blacklist' has been deprecated. Instead use 'property-disallowed-list'. See: https://github.com/stylelint/stylelint/blob/13.7.0/lib/rules/property-blacklist/README.md
Deprecation Warning: 'unit-blacklist' has been deprecated. Instead use 'unit-disallowed-list'. See: https://github.com/stylelint/stylelint/blob/13.7.0/lib/rules/unit-blacklist/README.md
```

ismail-syed:
Additional context, this is due to changes in [stylelint v13.7.0](https://github.com/stylelint/stylelint/blob/master/CHANGELOG.md#1370)

PR #188, opened 2020-10-06
Update deprecated rules

ismail-syed:
## Description

Fixes https://github.com/Shopify/web-configs/issues/187

## Type of change

- `@shopify/stylelint-plugin`
- [x]  Major: Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist


ismail-syed:
> Interesting they added a breaking change like this in a minor version?

I don't think it's actually a breaking change since their deprecation warnings, not runtime errors. That said, for our config we'll need to make it a breaking change since the `allowed/disallowed` rules are only available starting v13.7.0

PR #190, opened 2020-10-08
Expand the peerDependency on eslint to include 7

vsumner:
## Description
Expand the peerDependency for `eslint-plugin` to include `eslint@7`.

## Type of change

- [X] `@shopify/eslint-plugin` Patch: Bug (non-breaking change which fixes an issue)
expected)

## Checklist
I don't think this needs a changelog entry, but will add one if anyone feels that it should.

### DIFF
--- a/packages/eslint-plugin/package.json
+++ b/packages/eslint-plugin/package.json
@@ -30,7 +30,7 @@
     "typescript": "^3.9.2"
   },
   "peerDependencies": {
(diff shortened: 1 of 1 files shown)


BPScott (approved):
I can never remember if peer dep tweaks result in an update to yarn.lock files too. if not : shipit:

sanjanadesai27 (approved):

PR #198, opened 2020-11-27
Typescript naming rules

gmcgibbon:
## Description

Adds two naming rules:
- Ensure all generic type names start with a `T` prefix
- Ensure all interfaces _don't_ start with an `I` prefix

I think we could do with making a decision on these two naming conventions. These make sense to me for the following reasons:

#### The `T` prefix on generics:
- TypeScript is a Microsoft language, and `T` names and prefixes are in the [official docs](https://www.typescriptlang.org/docs/handbook/generics.html)
- Other MS languages like [C#](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.dictionary-2?view=net-5.0) use this standard (my background when it comes to typed languages)
- Makes it easier to identify when I type is a generic, and not a concrete type filling a generic's place
- It allows people coming from other MS languages to onboard quicker to TypeScript
- I can’t find an example in the handbook where it doesn't just use T or some other letter which is slightly disheartening, but using the prefix is a visual compromise between clear generic type indication and human readable naming

#### The lack of `I` prefix on interfaces:
- Again, the TypeScript handbook does not do this, and we currently don't either (from what I've seen)
- This seems like a very _enterprise_ rule that might be a little antiquated
- This probably doesn't need as much selling, but having a linting rule that explicitly complains when others do this will help make code consistent

## Type of change

- [x] `@shopify/esling-plugin` Minor: New feature (non-breaking change which adds functionality)

## Checklist

### DIFF
--- a/packages/eslint-plugin/lib/config/typescript.js
+++ b/packages/eslint-plugin/lib/config/typescript.js
@@ -77,6 +77,14 @@ module.exports = {
             format: ['PascalCase'],
             prefix: ['T'],
           },
+          {
+            selector: 'interface',
+            format: ['PascalCase'],
+            custom: {
+              regex: '^I[A-Z]',
+              match: false,
+            },
+          },
         ],
       }),
     },
--- a/packages/eslint-plugin/lib/config/typescript.js
+++ b/packages/eslint-plugin/lib/config/typescript.js
@@ -69,6 +69,15 @@ module.exports = {
         '@shopify/typescript/prefer-singular-enums': 'error',
         // Prefer buildClientSchema for schema building.
         '@shopify/typescript/prefer-build-client-schema': 'error',
+
+        '@typescript-eslint/naming-convention': [
+          'error',
+          {
+            selector: 'typeParameter',
+            format: ['PascalCase'],
+            prefix: ['T'],
+          },
+        ],
       }),
     },
   ],
--- a/packages/eslint-plugin/CHANGELOG.md
+++ b/packages/eslint-plugin/CHANGELOG.md
@@ -9,6 +9,7 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ### Changed
 
+- Updated TypeScript naming rules to ensure type name start with a "T" and interface names don't start with an "I" ([#198](https://github.com/Shopify/web-configs/pull/198))
 - Updated `eslint-config-prettier` dependency to v3.3.0 ([#202](https://github.com/Shopify/web-configs/pull/202))
 
 ## [39.0.3] - 2020-11-17


9at8:
Makes sense to me

atesgoral (approved):

kaelig:
Just to make sure I understand correctly the regexp, this still allows an interface to be called `Image` (starts with an `I`, followed by a lowercase character)?

packages/eslint-plugin/lib/config/typescript.js
```diff
@@ -69,6 +69,23 @@ module.exports = {
         '@shopify/typescript/prefer-singular-enums': 'error',
         // Prefer buildClientSchema for schema building.
         '@shopify/typescript/prefer-build-client-schema': 'error',
+
+        '@typescript-eslint/naming-convention': [
```
ismail-syed, line 73: Can we note this change in the changelog, please.

ismail-syed (approved):

ismail-syed:
> Ensure all generic type names start with a T prefix

This is 👍 with me.

> Ensure all interfaces don't start with an I prefix

I personally don't like interfaces that start with `I`, it just looks funny to me so I'm 👍 with this too.

atesgoral:
@kaelig Seems like it!

packages/eslint-plugin/lib/config/typescript.js
atesgoral: @ismail-syed I'll do it.

BPScott (approved):
I'm slightly throw as this seems too have two similar but subtly different ideas that I would have expected to be a bit more consistent:

- There is value in using a prefixed-based naming convention to tell the difference between a Generic Type and any other Type is good so we should enforce a prefix
- There is no value in using a prefix-based naming convention to tell the difference between a Type and a concrete value so we should ban using a prefix.


That's probably just me being contrarian though. I don't have a strong opinion on this "banning a thing we already basically never do" is fine by me :shipit:

atesgoral:
@BPScott I don't have a strong opinion on the matter, but if I were to try to justify the combination of these two new rules:

Perhaps duck typing can be seen as interface analogue and therefore we could say "don't prefix JS-native types, whether they're interfaces (duck-typed things) or concrete types implementing those interfaces". On the other hand, a generic type is a preprocessor thing that only belong to the TS realm and so we want to prefix generic types with "T" to make them stand out.

`¯\_(ツ)_/¯`

dahukish:
Not to go against what MS does, and this might be over simplifying things but I think prefixes in general lower the cognitive load when reading someone else's code. To that Effect I would say that we add both the `I` and the `T` or some other naming convention that avoids the assumption that my editor will pickup that it is an interface/type or that I will need to open multiple files. 🤷

dahukish:
@gmcgibbon we good to ship this?

PR #202, opened 2020-12-14
Bump eslint-plugin-prettier to v3.3.0

BPScott:
## Description

Bumps the dependency version of eslint-plugin-prettier to v3.3.0, which includes a mighty performance improvement for the prettier rule per https://github.com/prettier/eslint-plugin-prettier/pull/368

## Type of change

- [x] `@shopify/eslint-plugin` Patch: Bug (non-breaking change which fixes an issue)

## Checklist


ismail-syed (approved):
🙌 🙌 🙌

PR #206, opened 2021-02-10
Add babel transform runtime

vsumner:
## Description
This PR adds the ability to enable `transformRuntime` to add the babel plugin `@babel/plugin-transform-runtime`.

> When compiling TypeScript with only babel-loader, extra helpers are inlined in each source module. This normally adds some bulks to files, but `transform-runtime` converts these blocks into common helper imports that webpack can collate into shared assets.

## Type of change

- [X]  Minor: New feature (non-breaking change which adds functionality)

## Checklist

- [X] I have added a changelog entry, prefixed by the type of change noted above (Documentation fix and Test update does not need a changelog as we do not publish a new version for these changes)

### DIFF
--- a/packages/babel-preset/non-standard-plugins.js
+++ b/packages/babel-preset/non-standard-plugins.js
@@ -3,6 +3,13 @@ module.exports = function shopifyNonStandardPlugins(options = {}) {
     inlineEnv = false,
     typescript = false,
     transformRuntime = false,
+    transformRuntimeOptions = {
+      useESModules: false,
+      corejs: false,
+      regenerator: true,
+      helpers: true,
+      absoluteRuntime: false,
+    },
   } = options;
 
   const plugins = [require.resolve('@babel/plugin-syntax-dynamic-import')];
@@ -51,6 +58,7 @@ module.exports = function shopifyNonStandardPlugins(options = {}) {
       plugins.push([
         '@babel/plugin-transform-runtime',
         {
+          ...transformRuntimeOptions,
           version: require('@babel/plugin-transform-runtime/package.json')
             .version,
         },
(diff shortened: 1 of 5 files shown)


packages/babel-preset/non-standard-plugins.js
```diff
@@ -38,6 +42,20 @@ module.exports = function shopifyNonStandardPlugins(options = {}) {
...
+
+    if (transformRuntime) {
+      // When compiling TypeScript with only babel-loader, extra helpers are
+      // inlined in each source module. This normally adds some bulks to files,
+      // but `transform-runtime` converts these blocks into common helper imports
+      // that webpack can collate into shared assets.
+      plugins.push([
+        '@babel/plugin-transform-runtime',
+        {
+          version: require('@babel/plugin-transform-runtime/package.json')
```
GoodForOneFare, line 54: Doesn't have to happen in this PR, but I'd be keen to see how options that I missed work out:
* [`useESModules`](https://babeljs.io/docs/en/babel-plugin-transform-runtime#useesmodules)
* [`regenerator`](https://babeljs.io/docs/en/babel-plugin-transform-runtime#regenerator) - I think most Shopify apps have a global regenerator? So maybe this plugin doesn't have to do so much.
* [`corejs`](https://babeljs.io/docs/en/babel-plugin-transform-runtime#corejs) - maybe doesn't make sense until everything's on core-js@3

yarn.lock
```diff
@@ -173,6 +173,13 @@
   dependencies:
     "@babel/types" "^7.10.4"
 
+"@babel/helper-module-imports@^7.12.13":
```
GoodForOneFare, line 176: If possible, dedupe with the above `@babel/helper-module-imports@^7.10.4`.  Same goes for most of the added dependencies in this PR.

GoodForOneFare (approved):
👍

yarn.lock
vsumner: https://github.com/Shopify/web-configs/pull/207

packages/babel-preset/non-standard-plugins.js
vsumner: `useESModules` looks really promising

PR #208, opened 2021-02-10
Fix undefined check for transformReactConstantElements

vsumner:
## Description
This PR fixes an undefined check I made.

## Type of change

## Checklist


michenly (approved):
```
false || true = true
false ?? true = false
```

that's a mouthful 😂

PR #209, opened 2021-02-10
Much better way of setting options

vsumner:
## Description
This PR sets options in `babel-preset/react` the same way `babel-preset/web` does to be more consistent.

I 🎩 ed this on another repo.

## Type of change

## Checklist

### DIFF
--- a/packages/babel-preset/react.js
+++ b/packages/babel-preset/react.js
@@ -1,10 +1,12 @@
 module.exports = function shopifyReactPreset(api, options = {}) {
   const env = api.env();
 
-  const pragma = options.pragma || 'React.createElement';
-  const pragmaFrag = options.pragmaFrag || 'React.Fragment';
-  const transformReactConstantElements =
-    options.transformReactConstantElements ?? true;
+  const {
+    pragma = 'React.createElement',
+    pragmaFrag = 'React.Fragment',
+    transformReactConstantElements = true,
+  } = options;
+
   const plugins = [];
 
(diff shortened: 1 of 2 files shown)


michenly (approved):

packages/babel-preset/react.js
```diff
@@ -1,10 +1,12 @@
 module.exports = function shopifyReactPreset(api, options = {}) {
   const env = api.env();
 
-  const pragma = options.pragma || 'React.createElement';
-  const pragmaFrag = options.pragmaFrag || 'React.Fragment';
-  const transformReactConstantElements =
-    options.transformReactConstantElements ?? true;
+  const {
+    pragma = 'React.createElement',
+    pragmaFrag = 'React.Fragment',
```
michenly, line 6: the only way this "could" be a problem is these two are empty string and we are no longer catching it.

BPScott: Can we stick this destructuring straight in the parameters to dodge this intermediate value?

```
function shopifyReactPreset(
  api,
  {
    pragma = "React.createElement",
    pragmaFlag = "React.Fragment",
    transformReactConstantElements = true,
  } = {}
) {}

```

packages/babel-preset/react.js
```diff
@@ -1,10 +1,12 @@
...
+    transformReactConstantElements = true,
+  } = options;
```
dahukish, line 8: Forgive my ignorance but where do these actually get used. We set the defaults above but they are not used anywhere after?

PR #211, opened 2021-02-11
Fix the all.js test for the eslint-plugin

vsumner:
## Description
The `all.js` test works because it ignored the `.graphql` files included in the fixture and didn't apply the `.graphql` rules.
This PR updates the fixture and the `all.test.js` execution to include the graphql files.

## Type of change

- [X] `@shopify/eslint-plugin` Patch: Bug (non-breaking change which fixes an issue)

## Checklist

### DIFF
--- a/packages/eslint-plugin/tests/fixtures/all/app/graphql/Query.graphql
+++ b/packages/eslint-plugin/tests/fixtures/all/app/graphql/Query.graphql
@@ -1,5 +1,6 @@
-query LintErrorFooQuery {
+query AllTestQuery {
   foo {
-    DOES_NOT_EXIST
+    id
+    bar
   }
 }
--- a/packages/eslint-plugin/tests/lib/config/all.test.js
+++ b/packages/eslint-plugin/tests/lib/config/all.test.js
@@ -4,7 +4,9 @@ describe('config', () => {
   it('has valid plugins and requires', () => {
     expect(
       execESLint(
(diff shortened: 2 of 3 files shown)


atesgoral (approved):

packages/eslint-plugin/tests/fixtures/all/app/graphql/Query.graphql
```diff
@@ -1,5 +1,6 @@
-query LintErrorFooQuery {
+query AllTestQuery {
```
dahukish, line 1: 🏅

packages/eslint-plugin/tests/lib/config/all.test.js
```diff
@@ -4,7 +4,9 @@ describe('config', () => {
   it('has valid plugins and requires', () => {
     expect(
       execESLint(
-        `--config ${fixtureFile('all/.eslintrc.js')} ${fixtureFile('all')}`,
+        `--ext .js,.graphql --ignore-pattern "**/all/build/*" --config ${fixtureFile(
+          'all/.eslintrc.js',
+        )} ${fixtureFile('all')}`,
       ),
     ).toBe('');
   }, 8000);
```
dahukish, line 12: do we worry this timeout mike make our tests flaky at some point?

vsumner: I didn't add it but I assume it was expanded to stop it from being flakey in CI environments.

PR #212, opened 2021-02-12
Update @typescript-eslint/array-type to be compatible with @typescrip…

vsumner:
## Description
Update @typescript-eslint/array-type to be compatible with @typescript-eslint/ban-types

`array-simple` causes this error:
```
error: Array type using 'T[]' is forbidden for non-simple types. Use 'Array<T>' instead (@typescript-eslint/array-type) at packages/sewing-kit/src/tools/webpack/playground/index.ts:59:21:
  57 |     : playgroundForWorkspace();
  58 |
> 59 |   const filesToAdd: Promise<void>[] = [];
     |                     ^
  60 |
  61 |   if (force || !(await pathExists(playgroundFile))) {
  62 |     filesToAdd.push(writeFile(playgroundFile, format(playgroundContent)));
  ```
  
When the type gets updated to `Array<Promise<void>>` then the `@typescript-eslint/ban-types` is tripped.

I tested this against `sewing-kit`

## Type of change

- [x] `@shopify/eslint-plugin` Patch: Bug (non-breaking change which fixes an issue)

## Checklist

### DIFF
--- a/packages/eslint-plugin/CHANGELOG.md
+++ b/packages/eslint-plugin/CHANGELOG.md
@@ -9,7 +9,7 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ### Changed
 
-- Updated `eslint-plugin` to fix incompatibility between `@typescript-eslint/array-type` and `@typescript-eslint/ban-types` ([#212](https://github.com/Shopify/web-configs/pull/212))
+- Fix incompatibility between `@typescript-eslint/array-type` and `@typescript-eslint/ban-types` ([#212](https://github.com/Shopify/web-configs/pull/212))
 
 ## [40.0.0] - 2021-02-12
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: packages/eslint-plugin/CHANGELOG.md [hunk 1]
```diff
@@ -7,6 +7,10 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 <!-- ## [Unreleased] -->
 
+### Changed
+
+- Updated `eslint-plugin` to fix incompatibility between `@typescript-eslint/array-type` and `@typescript-eslint/ban-types` ([#212](https://github.com/Shopify/web-configs/pull/212))
```
gmcgibbon, line 12:
```suggestion
- Fix incompatibility between `@typescript-eslint/array-type` and `@typescript-eslint/ban-types` ([#212](https://github.com/Shopify/web-configs/pull/212))
```
The package name in the changelog is redundant since we keep one log per project, right?

gmcgibbon (approved):

PR #213, opened 2021-02-12
include jest-formatting rules in @shopify/eslint-plugin

IlanaB:
## Description

jest-formatting rules were added in https://github.com/Shopify/web-configs/pull/178 but they were never included in the default ruleset applied in `config/jest.js`. This PR includes the rules defined in that earlier PR in the ruleset applied when consumers extend `plugin:@shopify/jest` in their `eslintrc`

## Type of change
- [x] `@shopify/eslint-plugin` Minor: New feature (non-breaking change which adds functionality)

## Checklist


BPScott (approved):

BPScott:
Not ran but looks good to me!

ryanwilsonperkin (approved):

PR #215, opened 2021-02-18
Move to Github actions for CI

BPScott:
## Description

Moving from travis to GH actions for CI


BPScott:
I'm an impatient fool, and assumed not seeing an action run in this PR was due to the workflow not being merged into main when it was actually because I can't write valid yaml.

Here's the fix that gets CI working: https://github.com/Shopify/web-configs/commit/6341b899767e4d756944233b7b80d0764ae84ee9

PR #231, opened 2021-04-13
Common babel plugin

vsumner:
## Description
This PR adds an experimental `common` babel-preset that supports node and web presets.

## Type of change

- [X]  `babel-preset` (non-breaking change which adds functionality)

## Checklist


packages/babel-preset/common.js
```diff
@@ -0,0 +1,100 @@
...
+    ],
+  ];
+
+  const plugins = [
+    ...nonStandardPlugins(options),
+    isNode && require.resolve('@babel/plugin-proposal-dynamic-import'),
+    isNode && require.resolve('@babel/plugin-transform-modules-commonjs'),
+    typescript && require.resolve('@babel/preset-typescript'),
+    // polyfill for import statements in webpack for browsers that don't support promises
+    require.resolve('@babel/plugin-syntax-dynamic-import'),
```
kaelig, line 55: Is this still useful? Don't we support browsers that support promises?

packages/babel-preset/common.js
```diff
@@ -0,0 +1,100 @@
...
+    // proposal-decorators must go before proposal-class-properties.
+    // Typscript implements the stage 1 version of decorators, which is the
```
kaelig, line 57:
```suggestion
    // TypeScript implements the stage 1 version of decorators, which is the
```

Also: what version of TypeScript?

packages/babel-preset/common.js
```diff
@@ -0,0 +1,100 @@
...
+    // Polyfills the runtime needed for async/await, generators, and friends
+    // https://babeljs.io/docs/en/babel-plugin-transform-runtime
+    transformRuntime && [
+      '@babel/plugin-transform-runtime',
+      {
+        ...transformRuntimeOptions,
+        version: require('@babel/runtime/package.json').version,
+      },
+    ],
+    !typescript && require.resolve('@babel/plugin-proposal-class-properties'),
```
kaelig, line 96: Would it make sense to only mention this plugin once (and delete the prior mention above, lines 66-69)?

```suggestion
    [
      require.resolve('@babel/plugin-proposal-class-properties'),
      // Enable loose mode in TypeScript to use assignment instead of defineProperty
      {loose: typescript},
    ],
```

packages/babel-preset/common.js
vsumner: Good point. It most likely isn't useful anymore.

BPScott: I think `@babel/plugin-proposal-dynamic-import`, `@babel/plugin-syntax-dynamic-import` and `@babel/plugin-transform-modules-commonjs` are all provided by preset-env in 7.13 (probably added in earlier versions) so we don't need to explicitly state them anymore.


Using blame in https://github.com/babel/babel/blob/main/packages/babel-preset-env/src/available-plugins.js be useful for researching this

packages/babel-preset/common.js
```diff
@@ -0,0 +1,100 @@
...
+  const isNode = 'node' in targets;
+
+  const presets = [
+    [
+      require.resolve('@babel/preset-env'),
+      {
+        modules,
+        useBuiltIns,
+        corejs,
+        targets,
```
BPScott, line 42: Babel 7.13 shipped [top-level targets](https://babeljs.io/blog/2021/02/22/7.13.0#top-level-targets-option-12189httpsgithubcombabelbabelpull12189-rfchttpsgithubcombabelrfcspull2), which means we don't need to drill them into preset-env anymore. i'd like to see us stop passing this into preset-env and instead use the top-level option.

packages/babel-preset/common.js
vsumner: Yes this is still required. Decorator support is only experimental in typescript an not supported by babel-typescript.

packages/babel-preset/common.js
```diff
@@ -0,0 +1,100 @@
...
+    typescript && [
+      require.resolve('@babel/plugin-proposal-decorators'),
+      {legacy: true},
+    ],
+    // Enable loose mode to use assignment instead of defineProperty
+    typescript && [
+      require.resolve('@babel/plugin-proposal-class-properties'),
+      {loose: true},
+    ],
+    // Adds Numeric Separators
```
BPScott, line 70: At the time this was originally written numeric-separator was stage 3 and thus not part of preset-env's defaults. now it's part of preset env. and unfortunately though it can't be removed.

Numeric separators now falls into the same trap as nullish-coalescing and optional-chaning - handed by preset-env but not supported by acorn v6 ([support added in 7.4](https://github.com/acornjs/acorn/blob/master/acorn/CHANGELOG.md#740-2020-08-03)) so thus we always have to transform them, so it'd be worth putting this under comment below.

packages/babel-preset/common.js
```diff
@@ -0,0 +1,100 @@
+const nonStandardPlugins = require('./non-standard-plugins');
+
+module.exports = function shopifyCommonPreset(api, options = {}) {
+  const {
+    targets = {
```
BPScott, line 5: I think we should push towards only browserslist strings as targets rather than using the object notation. `current node` is a valid browserlist target.

Pending comments below I think we remove the need for `isNode` and thus for this to be an object

packages/babel-preset/common.js
```diff
@@ -0,0 +1,100 @@
...
+        corejs,
+        targets,
+        debug,
+        bugfixes: true,
+      },
+    ],
+  ];
+
+  const plugins = [
+    ...nonStandardPlugins(options),
```
BPScott, line 50: this seems to duplicate most of the inline definitions you've provided here. Can it be removed?

packages/babel-preset/common.js
```diff
@@ -0,0 +1,100 @@
...
+      corejs: false,
+      helpers: true,
+      // By default, babel assumes babel/runtime version 7.0.0-beta.0,
+      // explicitly resolving to match the provided helper functions.
+      // https://github.com/babel/babel/issues/10261
+      version: require('@babel/runtime/package.json').version,
+      regenerator: true,
+      // https://babeljs.io/docs/en/babel-plugin-transform-runtime#useesmodules
+      // We should turn this on once the lowest version of Node LTS
+      // supports ES Modules.
```
BPScott, line 24:
```suggestion
      // supports ES Modules and exports maps.
```

packages/babel-preset/common.js
kaelig: My question was more around whether we can mention a specific version of TypeScript and if this could break when used with future versions of TypeScript.

packages/babel-preset/common.js
vsumner: Thanks. I missed it, this was my intention by moving its contents into here.

packages/babel-preset/common.js
```diff
@@ -0,0 +1,81 @@
...
+    // But they aren't yet supported in webpack 4 because of missing support
+    // in acorn v6 (support is in acorn v7, which is used in webpack v5).
+    // So we want to always transpile this synax away
+    // See https://github.com/webpack/webpack/issues/10227
+    // Can be removed once we drop support for webpack v4 (or these features
+    // are backported to acorn v6)
+    typescript && require.resolve('@babel/plugin-proposal-numeric-separator'),
+    typescript && [
+      require.resolve('@babel/plugin-proposal-nullish-coalescing-operator'),
+      {loose: true},
```
BPScott, line 63: If we're in the process of making a breaking from existing behaviour, what do you think about removing `{loose: true}` for nullish-coalescing-operator and optional-chaining so that we know we're adhering to the strict behaviour  - and thus we behave in the same way as native code and `preset-env` as we don't pass the loose option into that. 

https://babeljs.io/docs/en/babel-plugin-proposal-nullish-coalescing-operator#loose
https://babeljs.io/docs/en/babel-plugin-proposal-optional-chaining#loose

packages/babel-preset/CHANGELOG.md
```diff
@@ -7,6 +7,14 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 <!-- ## Unreleased -->
 
+### Added
+
+- Created a `common` babel-preset [[#231](https://github.com/Shopify/web-foundation/pull/231)]
```
BPScott, line 12: Do we want to make a bigger deal out of this being experimental for now?

I'm assuming this is a minor release for now, and in a few weeks once it proves itself we will cut a major version that exposes the common preset in `index.js` and remove the web/node presets.

BPScott (approved):
Code as-is looks great!

Dropped a few nitpick comments inline about potential for using strict instead of loose transforms, and wondering about docs but they can be ignored if desired.

I wonder if we want to go even further and merge the react preset into this this too? Having a  `react` and `reactOptions` keys in a similar way to `transformRuntime`/`transformRuntimeOptions`. Thats proobably overkill for now though.


I'm assuming this is a minor version release, then in the future we'll remove the web/node presets and expose "common" in index.js - so users specify `@shopify/babel-preset` instead of `@shopiify/babel-preset/common`.

packages/babel-preset/common.js
vsumner: Good call out. Removed.

packages/babel-preset/README.md
```diff
@@ -134,3 +134,131 @@ As noted above, you can include multiple of these presets together. Some common
...
+
+`boolean`, defaults to `false`
+
+This option when `true` will enable the `@babel/plugin-transform-runtime` plugin.
+
+[Documentation](https://babeljs.io/docs/en/babel-plugin-transform-runtime)
+
+#### `transformRuntimeOptions`
+
+This option configures the `@babel/plugin-transform-runtime` plugin when the `transformRuntime` option has been enabled.
```
BPScott, line 205: Do we need to explain each of these option in our own words? Could we get away with linking to https://babeljs.io/docs/en/babel-plugin-transform-runtime#options and noting any cases where our defaults differ from babel.

Same goes for the reactPresetOptions too.

There's probably something we can do with the other options to say "same as preset-env" too.

packages/babel-preset/common.js
```diff
@@ -0,0 +1,109 @@
+module.exports = function shopifyCommonPreset(api, options = {}) {
+  const {
```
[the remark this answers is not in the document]
vsumner: Thanks. I forgot I'm not passing options to anything else anymore.

PR #236, opened 2021-04-21
Add option for enabling plugins for dynamic imports on node webpack

vsumner:
## Description
This PR adds a new options to the common preset. `nodeWebpack` which enables the `'@babel/plugin-proposal-dynamic-import'` and the `'@babel/plugin-transform-modules-commonjs'` plugins.

## Type of change

- [x] `@shopify/babel-preset` Minor: New feature (non-breaking change which adds functionality)

## Checklist


michenly (approved):
Did not 🎩 but I think the code make sense :)

vsumner:
Turns out I don't need this. The plugins are provided by preset-env if we set the right settings.

PR #224, opened 2021-04-01
Update stylelint

vsumner:
## Description
Update `stylelint`, `stylelint-prettier`, and `stylelint-scss`

https://github.com/kristerkari/stylelint-scss/releases
https://github.com/prettier/stylelint-prettier/blob/master/CHANGELOG.md
https://github.com/stylelint/stylelint/blob/master/CHANGELOG.md#13120

## Type of change

- [ ]  Patch: Bug (non-breaking change which fixes an issue)
- [X] `@shopify/stylelint-plugin` Minor: New feature (non-breaking change which adds functionality)
- [ ]  Major: Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

- [x] I have added a changelog entry, prefixed by the type of change noted above (Documentation fix and Test update does not need a changelog as we do not publish a new version for these changes)

PR #237, opened 2021-04-22
Remove non-existent rule and bump stylelint peer dep

BPScott:
## Description

https://github.com/Shopify/web-configs/pull/224 added some new stylelint rules. However `no-invalid-position-at-import-rule` isn't actually released yet per [stylelint's changelog](https://github.com/stylelint/stylelint/blob/master/CHANGELOG.md), and thus using the config results in an "Unknown rule" error.

Also bump the stylelint peer dependency to ensure consumers use a stylelint version that has the other rules that we specify.

## Type of change

- [x] stylelint-config Patch: Bug (non-breaking change which fixes an issue)

## Checklist


BPScott:
Things be breaking, nobody's around YOLO Merge time!

michenly (approved):
I didnt 🎩 but ok with the code

PR #243, opened 2021-05-05
Upgrade to merge 2.1.1

melodyhabbouche:
## Type of change

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,5 +27,8 @@
     "plop": "^2.6.0",
     "prettier": "^2.0.5",
     "watchpack": "^1.6.1"
+  },
+  "dependencies": {
+    "merge": "^2.1.1"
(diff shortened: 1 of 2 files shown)


kaelig:
Thank you for the contribution! I see that the dependency was added at the root level, which won't update the dependencies in `@shopify/eslint-plugin` and `@shopify/stylelint-plugin` (they're still using `merge@1.2.1`).

There are multiple ways of doing this, but the way I'd go about it is by using `yarn workspace` ([documentation](https://classic.yarnpkg.com/en/docs/cli/workspace/)) to remove and then re-add the `merge` dependency in these workspaces.

melodyhabbouche:
Closing because Victor handled in another PR!

PR #247, opened 2021-05-10
@shopify/eslint-plugin: merge typescript naming convention rules

brendanrygus:
## Description
While upgrading to [v40.2.2](https://github.com/Shopify/web-configs/releases/tag/%40shopify%2Feslint-plugin%4040.2.2), we noticed some changes in the  `@typescript-eslint/naming-convention` rule.
Some of the recent changes around [enforcing T prefixes for generic types](https://github.com/Shopify/web-configs/pull/198) might have loosened up other checks.

These previously invalid declarations all became valid in ts/tsx files:
```
type myCamelCasedType = 'ok'
const my_snake_cased_var = 'ok, ruby'
const object = {
  my_backend_enum_key: 'yup cool',
  MY_SCREAMING_KEYS: 'whatever'
}
```

## Type of change
This PR merges the two `@typescript-eslint/naming-convention` rules together to restore the previous config.

- [x] @shopify/eslint-plugin Patch: Bug (non-breaking change which fixes an issue)

## Checklist


packages/eslint-plugin/lib/config/rules/typescript.js
```diff
@@ -131,6 +131,19 @@ module.exports = {
       selector: 'typeLike',
```
brendanrygus, line 131: I used the `--print-config` CLI arg to log out the eslint rules our repo used before the recent changes

I confirmed that the rules for `typescript-eslint/naming-convention` matched those in this file, so merging the two rule sets together seems like our best bet
```
    "@typescript-eslint/naming-convention": [
      "error",
      {
        "selector": "default",
        "format": [
          "camelCase",
          "PascalCase",
          "UPPER_CASE"
        ],
        "leadingUnderscore": "allow",
        "trailingUnderscore": "allow"
      },
      {
        "selector": "default",
        "filter": {
          "match": true,
          "regex": "^(__|UNSAFE_).+$"
        },
        "format": null
      },
      {
        "selector": "typeLike",
        "format": [
          "PascalCase"
        ]
      }
    ],
```

packages/eslint-plugin/lib/config/rules/typescript.js
```diff
@@ -131,6 +131,19 @@ module.exports = {
...
       format: ['PascalCase'],
     },
+    {
+      selector: 'typeParameter',
+      format: ['PascalCase'],
+      prefix: ['T'],
+    },
+    {
+      selector: 'interface',
```
dahukish, line 140: Makes sense to me.

packages/eslint-plugin/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 <!-- ## Unreleased -->
 
+### Changed
```
brendanrygus, line 10: Thanks @vsumner !

PR #251, opened 2021-05-21
WIP - Add require-autocomplete lint rule

andrewmcgov:
## Type of change

## Checklist


packages/eslint-plugin/lib/rules/react-require-autocomplete.js
```diff
@@ -0,0 +1,137 @@
...
+  'password',
+  'range',
+  'search',
+  'tel',
+  'text',
+  'time',
+  'url',
+  'week',
+];
+const autocompleteElementNames = ['input', 'TextField'];
```
andrewmcgov, line 19: If we have other components that some libraries may want the rule applied to, I _think_ we could make this a setting.

packages/eslint-plugin/lib/config/rules/shopify.js
```diff
@@ -45,6 +45,8 @@ module.exports = {
   '@shopify/react-no-multiple-render-methods': 'off',
   // Prefer all non-React-specific members be marked private in React class components.
   '@shopify/react-prefer-private-members': 'off',
+  // Require input elements to have autocomplete values
+  '@shopify/react-require-autocomplete': 'off',
```
andrewmcgov, line 49: Does this make the rule disabled by default, allowing apps to opt in on their own time?

packages/eslint-plugin/tests/lib/rules/react-require-autocomplete.test.js
```diff
@@ -0,0 +1,62 @@
...
+        return <TextField type="email" {...fieldProps} />;
+      }`,
+    },
+    {
+      code: `
+      function MyComponent({fieldProps}) {
+        return <TextField type="text" {...fieldProps} />;
+      }`,
+    },
+  ],
```
Tom-Bonnike, line 46: I think we could add two more test cases:
-  for an `input` type on which `autoComplete` is not allowed (e.g. `<input type="checkbox" />`)
- a test where it doesn’t warn about components other than `input` and `TextField`

packages/eslint-plugin/lib/config/react.js
```diff
@@ -26,6 +26,7 @@ module.exports = {
       '@shopify/react-initialize-state': 'error',
       '@shopify/react-no-multiple-render-methods': 'error',
       '@shopify/react-prefer-private-members': 'error',
+      '@shopify/react-require-autocomplete': 'error',
```
Tom-Bonnike, line 29: I’m guessing you thought this through already but… could there be any case where we wouldn’t want an `autoComplete` on one of these input types? Asking because maybe it should simply be a `warn`.

packages/eslint-plugin/lib/rules/react-require-autocomplete.js
```diff
@@ -0,0 +1,137 @@
...
+
+      context.report(
+        node,
+        `Input elements of type '${inputType}' must have an autoComplete attribute`,
+      );
+    },
+  }),
+};
+
+function findAttrValue(attributes, name, context) {
```
Tom-Bonnike, line 67: Interesting, I’d recommend using a module like [jsx-ast-utils](https://www.npmjs.com/package/jsx-ast-utils) so we don’t have to implement these utils ourselves?

Tom-Bonnike (commented):
Super cool, good work! Your commit still mentions it‘s a WIP you may want to amend that 😜

vsumner (approved):
Great work! Thank you for providing tests and the detailed documentation of this new rule.

I tested this against `Shopify/web` and because the rule is set to `error` for `react` it is indeed on by default.

There were 54 errors and they seem to be legitimate.

```
error: Input elements of type 'text' must have an autoComplete attribute (@shopify/react-require-autocomplete) at app/components/SkeletonIndex/SkeletonIndex.tsx:34:15:
  32 |             <SkeletonTabs />
  33 |             <Card.Section>
> 34 |               <TextField
     |               ^
  35 |                 connectedLeft={filterButton}
  36 |                 label=""
  37 |                 labelHidden


error: Input elements of type 'text' must have an autoComplete attribute (@shopify/react-require-autocomplete) at app/components/SsnField/SsnField.tsx:41:5:
  39 |
  40 |   return (
> 41 |     <TextField
     |     ^
  42 |       value={format(value)}
  43 |       label={label}
  44 |       helpText={helpText}
  ```

I think minor is safe since this is a new feature but not a breaking change.

File: packages/eslint-plugin/lib/rules/react-require-autocomplete.js (hunk 1 above)
Comment (Tom-Bonnike) [reply packages/eslint-plugin/lib/rules/react-require-autocomplete.js:19]: Ha, I was curious about whether we had the `autocomplete-valid` rule and it seems like we apply it on the [<Select> component](https://github.com/Shopify/web-configs/blob/9852a937a6b51edc60215dcc34c1b943cb648fab/packages/eslint-plugin/lib/config/rules/jsx-a11y.js#L75-L80) too!

**Comment by andrewmcgov** [comment]
Thanks @vsumner and @Tom-Bonnike! 

I've update the PR with your feedback and added a changelog entry. 

- We're going to leave this as `error` by default and let individual repos turn it off if they want to. 
- Thanks to Tom's suggestion I used a couple helper functions from [jsx-ast-utils](https://www.npmjs.com/package/jsx-ast-utils). This lead to me being able to delete most of the code I had written for the rule 😅. I ran the same test that Victor ran on web and got 53 errors instead of 54, which is fine.
- I added a setting to the rule for repo's to add elements to have the rule applied to. For example to test in web I added the following to the `.eslintrc.js` file: 

```js
'@shopify/react-require-autocomplete': [
  'error',
  {
    inputComponents: ['TextField'],
  },
],
```

Let me know if I am missing anything else!

packages/eslint-plugin/docs/rules/react-require-autocomplete.md
```diff
@@ -0,0 +1,31 @@
+# Require inputs have have autocomplete attributes when necessary
+
+The HTML `autoComplete` helps users to complete filling in forms by using data stored in the browser. This is particularly useful for people with motor disabilities or cognitive impairment who may have difficulties filling out forms online.
```
BPScott, line 3:
```suggestion
The HTML `autocomplete` attribute helps users to complete filling in forms by using data stored in the browser. This is particularly useful for people with motor disabilities or cognitive impairment who may have difficulties filling out forms online.
```

packages/eslint-plugin/lib/config/rules/shopify.js
BPScott: Correct.

packages/eslint-plugin/lib/rules/react-require-autocomplete.js
```diff
@@ -0,0 +1,70 @@
...
+      // input elements without a type attribute are treated as text
+      // https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Input#input_types
+      const inputType =
+        getPropValue(getProp(node.attributes, 'type')) || 'text';
+
+      if (!autoCompleteInputTypes.includes(inputType)) {
+        return;
+      }
+
+      const hasAutoCompleteProp = hasProp(node.attributes, 'autoComplete');
```
BPScott, line 58: Should attributes be lowercase rather than camel case? `autocomplete` rather than `autoComplete`?

packages/eslint-plugin/lib/utilities/index.js
```diff
@@ -147,6 +147,7 @@ module.exports = {
   polarisComponentFromJSX,
   getRootObject,
   docsUrl,
+  findDefinition,
```
BPScott, line 150: Where is this used, can it be removed? It does not seem to be a top-level function in this file so iI'd be suprised if this even works.

BPScott (commented):
Code is mostly looking good. Added a few questions inline, mainly around capitalisation of attributes - i thought they'd always be lowercase.

packages/eslint-plugin/docs/rules/react-require-autocomplete.md
andrewmcgov: I'm using `autoComplete` with an uppercase `c` here because the rule is targeted towards React, and that the supported attribute has the uppercase letter: https://reactjs.org/docs/dom-elements.html#all-supported-html-attributes.

Do you think I should change the value here in the docs?

packages/eslint-plugin/CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 <!-- ## Unreleased -->
```
BPScott, line 8: Uncomment the heading, and add a subheading now there is changelog content for a new release

```suggestion
## Unreleased

### Changed
```

BPScott (approved):
Nice! Thanks for doing this, and thanks for teaching me `autoComplete` is one of those react uses different conventions edge-cases.

One nitpick about changelog formatting to fixup then you can go squash and merge!

PR #252, opened 2021-05-25
Remove separate configs in favour of common

vsumner:
## Description
This PR removes the separate babel configs and replaces them with the common config.

Todo:
- Refactor documentation/readme

## Type of change

- [X] `@shopify/babel-preset`  Major: Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist


babel.config.js
```diff
@@ -1,3 +1,3 @@
 module.exports = {
-  presets: [['@shopify/babel-preset/node', {modules: 'commonjs'}]],
+  presets: [['@shopify/babel-preset/common', {modules: 'commonjs'}]],
```
BPScott, line 2: You're no longer specifying a target for babel to compile to (as previously the node preset defined it) so it will now compile everything down to es5, which is way overkill.

At a guess this is why you had to add regenerator-runtime deps - as tests failed because the transform down to es5 transpiled async code to use generators, which depends on regenerator-runtime.

Targeting `current node` should stop that transform from occurring - the node versions in dev.yml / CI support `async` nativly.

packages/babel-preset/index.js
```diff
@@ -1 +1 @@
-module.exports = require('./web');
+module.exports = require('./common');
```
BPScott, line 1: As we're making a breaking change anyway could we move  the contents of common.js into index.js? and then we can reference the preset `@shopify/babel-preset` rather than `@shopify/babel-preset/common`

packages/babel-preset/README.md
```diff
@@ -25,118 +25,11 @@ Then, in your Babel configuration (which should be under the `babel` key of your
...
-    ]
-  }
-}
-```
-
-## Experimental Presets
-
 ## `@shopify/babel-preset/common`
 
 The `@shopify/babel-preset/common` preset is designed to be a common import that offers more options. It is a combination of the `web`, `node`, and `react` presets all in one.
```
BPScott, line 35: Gonna need to update this bit of the doc, as this is saying "this is a bit like that thing that no longer exists"

PR #259, opened 2021-06-11
Update to lerna 4

dahukish:
## Description

Update this repo to use `lerna@4.0.0`

## Type of change

- [x]  Major: Breaking change 

## Checklist


.github/workflows/ci.yml
```diff
@@ -17,7 +17,7 @@ jobs:
 
       - uses: actions/setup-node@v2
         with:
-          node-version: '10.17.0'
+          node-version: '10.18.0'
```
BPScott, line 20: 10 is eol should we drag this up to 12.14.0 to match quilt?

dahukish: yeah... I was toying with doing there here or in a new PR.

BPScott (approved):

PR #263, opened 2021-06-15
Add resolutions for lerna transitive deps

dahukish:
## Description

Trying to get around lerna 4 not updating their old ass dependencies

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,5 +27,9 @@
     "plop": "^2.6.0",
     "prettier": "^2.0.5",
     "watchpack": "^2.2.0"
+  },
+  "resolutions": {
+    "trim-newlines": "3.0.1",
+    "normalize-url": "4.5.1"
   }
 }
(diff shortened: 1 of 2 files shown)


BPScott (commented):
I see resolutions as last resort. They're inherently fragile as hell because they override dependency requirements from transitive deps. We're basically saying "we know your dependencies better than you do" which is a very bold assumption that should be taken with great care IMO.

I don't think "a dependency-only tool is vulnerable to ReDos if given maliciously crafted strings, but in reality that will never happen because we control the input to lerna" is a strong enough reason to require using resolutions.

BPScott (commented):
looks like we can update normalize-url to a newer version by updating parse-url and postcss-normalize-url

dahukish:
> I see resolutions as last resort. They're inherently fragile as hell because they override dependency requirements from transitive deps. We're basically saying "we know your dependencies better than you do" which is a very bold assumption that should be taken with great care IMO.
> 
> I don't think "a dependency-only tool is vulnerable to ReDos if given maliciously crafted strings, but in reality that will never happen because we control the input to lerna" is a strong enough reason to require using resolutions.

The only option to not use resolutions for `trim-newlines` would be to update meow in lerna.

PR #264, opened 2021-06-21
[Eslint] jsx-ally | Removing jsx autocomplete valid rule

gcornhill:
## Description

We have a rule coming soon to require autocomplete on all inputs. This means there will be a lot of `autocomplete="off/nope"`. However `nope` is not valid, but due to longstanding chrome bug `nope` is the random string that will actually turn it off for Chrome. See [notes here](https://development.shopify.io/engineering/developing_at_Shopify/accessibility/forms/autocomplete#Handling_autocomplete_quot_off_quot_).

One the required rules go out, we will look to make our own valid rule like the above but allow `nope`.

## Type of change

- [x]  Patch: Bug (non-breaking change which fixes an issue)

## Checklist

### DIFF
--- a/packages/eslint-plugin/CHANGELOG.md
+++ b/packages/eslint-plugin/CHANGELOG.md
@@ -5,7 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-<!-- ## Unreleased -->
+## Unreleased
+
+### Changed
+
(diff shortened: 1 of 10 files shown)


andrewmcgov (approved):

BPScott:
should we wait for https://github.com/Shopify/web-configs/pull/251 to be merged and then this PR can diisable the old and enable the new in one go?

packages/eslint-plugin/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 <!-- ## Unreleased -->
```
BPScott, line 8: Convention is to uncomment the unreleased heading when there is content underneath it

```suggestion
## Unreleased
```

PR #267, opened 2021-07-07
Refactor @shopify/postcss-plugin for version 8

vsumner:
## Description
This PR refactors `@shopify/postcss-plugin` to match the new PostCSS 8 api. It also removes features that are unused.

## Type of change

- [x] `@shopify/postcss-plugin`  Major: Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

### DIFF
--- a/packages/postcss-plugin/index.js
+++ b/packages/postcss-plugin/index.js
@@ -2,15 +2,13 @@ const postcss = require('postcss');
 
 const features = require('./features');
 
-const plugin = postcss.plugin('@shopify/postcss-plugin', (options = {}) => {
-  const processor = postcss();
-
+module.exports = (options = {}) => {
+  const postCssPlugins = [];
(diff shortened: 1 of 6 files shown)


dahukish (approved):

packages/postcss-plugin/index.js
```diff
@@ -1,36 +1,14 @@
...
-        ],
-      }),
-    );
-  }
-
-  return processor;
-});
-
-module.exports = plugin;
+module.exports = ({calcOptions = {}, autopprefixerOptions = {}}) => {
```
BPScott, line 3: My fat fingers made it into your code! one too many p's in `autopprefixerOptions` there :D

Fix that then, I think this is good to go.

BPScott (approved):
🎉

Issue #246, opened 2021-05-05
Investigate using assumptions for babel-preset


vsumner:
## Overview

See @BPScott comment [here](https://github.com/Shopify/web-configs/pull/245#issuecomment-832930581).

BPScott:
https://babeljs.io/docs/en/babel-plugin-proposal-private-methods#options docs have been improved!

it suggests instead of using the `loose` option:

> ⚠️ Consider migrating to the top level `privateFieldsAsProperties` assumption.

This should allow us to remove explicitly setting the loose option on multiple plugins, and possibly removing the usage of those plugins entirely if they are already enabled as part of preset-env

PR #269, opened 2021-07-07
Add assumptions

dahukish:
## Description

closes: https://github.com/Shopify/web-configs/issues/246

## Type of change

- [x] `babel-preset` Patch.

## 🎩 

I tested this in global-nav using yalc and the latest sewing-kit-next. When running `yarn run test` with a fresh `node_modules` and no skn cache without the yalc'ed version of `babel-preset` you will get something like this:
```
Though the "loose" option was set to "false" in your @babel/preset-env config, it will not be used for @babel/plugin-proposal-private-methods since the "loose" mode option was set to "true" for @babel/plugin-proposal-class-properties.
The "loose" option must be the same for @babel/plugin-proposal-class-properties, @babel/plugin-proposal-private-methods and @babel/plugin-proposal-private-property-in-object (when they are enabled): you can silence this warning by explicitly adding
	["@babel/plugin-proposal-private-methods", { "loose": true }]
to the "plugins" section of your Babel config.
```
When running with a fresh `node_modules` and no cache AND the yalc'ed preset with the assumptions from this PR, no warnings were present in the console.

### Staging
Tested in org-admin here: https://staging-5.shopify.plus/plus-product-test-one/overview. SSR still works from what I can see.
Tested in the store admin here https://rules-test-store-4.myshopify.com/admin
Tested in shrink-ray staging https://shrink-ray-staging.shopifycloud.com/

`<div id="app">...` with contents is present when viewing source for both of the above.

## Checklist


packages/babel-preset/index.js
```diff
@@ -114,5 +109,10 @@ module.exports = function shopifyCommonPreset(
       require.resolve('@babel/plugin-transform-react-constant-elements'),
   ].filter(Boolean);
 
-  return {presets, plugins};
+  const assumptions = {
```
BPScott, line 112:
```suggestion
  // These assumptions are required because we use `@babel/plugin-proposal-decorators` in legacy mode
  // for compatibility with typescript
  // see https://babeljs.io/docs/en/babel-plugin-proposal-decorators#legacy
  // see https://babeljs.io/docs/en/babel-plugin-proposal-class-properties#loose
  // https://babeljs.io/docs/en/babel-plugin-proposal-private-methods#loose
  const assumptions = {
```

packages/babel-preset/index.js
```diff
@@ -73,15 +73,10 @@ module.exports = function shopifyCommonPreset(
     // Enable loose mode to use assignment instead of defineProperty when typescript is enabled
     // class-properties are handled by preset-env
     // But when using typescript we need to transpile them in loose mode to support proposal-decorators's legacy mode
-    typescript && [
-      require.resolve('@babel/plugin-proposal-class-properties'),
-      {loose: true},
-    ],
+    typescript && require.resolve('@babel/plugin-proposal-class-properties'),
```
BPScott, line 76: https://babeljs.io/blog/2021/04/29/7.14.0#new-class-features-enabled-by-default suggests plugin-proposal-class-properties and plugin-proposal-private-methods are now both enabled by preset-env. Thus we might be able stop referencing manually including these plugins entirely

However https://babeljs.io/docs/en/babel-plugin-proposal-decorators#note-compatibility-with-babelplugin-proposal-class-properties suggest that if you are including your plugins manually and using `@babel/plugin-proposal-class-properties`, then `@babel/plugin-proposal-decorators` must come before `@babel/plugin-proposal-class-properties`.

How do those two statements interact? Are we able to remove these two explicit plugin references (deleting this and the next line, or does the usage of `@babel/plugin-proposal-decorators` force us to always include these lines)

---


I wonder if we should add a new option `decorators` and make support for them (and the enabling of the assumptions) opt-in. They're non-standard, stuck at stage-2 since forever, experimental, and force transpilation of other standard features in a non-standard way. That doesn't feel like an approach that we should be advocating for general adoption by enabling it by default.

packages/babel-preset/index.js
```diff
@@ -73,15 +73,10 @@ module.exports = function shopifyCommonPreset(
     // Enable loose mode to use assignment instead of defineProperty when typescript is enabled
```
BPScott, line 73: These comments need to be updated as we're not controling this by using loose options anymore

packages/babel-preset/index.js
```diff
@@ -114,5 +109,10 @@ module.exports = function shopifyCommonPreset(
       require.resolve('@babel/plugin-transform-react-constant-elements'),
   ].filter(Boolean);
 
-  return {presets, plugins};
+  const assumptions = {
+    setPublicClassFields: true,
```
BPScott, line 113: I think these should only be set when `typescript` is true, as before we only enabled loose mode when typescript was true

BPScott (commented):
we need to update /adjust comments to explain what is happening - we're twiddling adjustments /plugins based on the presence of the decorators proposal

Added a question inline about if we still need to be explicit in adding these proposals that are now part of preset-env.

babel.config.js
```diff
@@ -1,4 +1,4 @@
 module.exports = {
   targets: 'current node',
-  presets: [['@shopify/babel-preset', {modules: 'commonjs'}]],
+  presets: [['@shopify/babel-preset', {modules: 'commonjs', typescript: true}]],
```
dahukish, line 3: We use typescript in our tests so I added this.

package.json
```diff
@@ -19,10 +19,11 @@
   },
   "devDependencies": {
     "@shopify/tophat": "^0.0.4",
+    "@types/jest": "^26.0.24",    
     "eslint": "^7.30.0",
     "fs-extra": "^9.0.0",
     "glob": "^7.1.6",
-    "jest": "^26.6.3",
+    "jest": "^27.0.6",
```
dahukish, line 26: new jest!

packages/babel-preset/index.js
```diff
@@ -28,6 +28,7 @@ module.exports = function shopifyCommonPreset(
     },
     transformReactConstantElements = false,
     isWebpack5 = false,
+    legacyDecorators = false,
```
dahukish, line 31: Decoupled this from typescript as this is specifically about decorators now.

packages/babel-preset/index.js
```diff
@@ -64,41 +65,26 @@ module.exports = function shopifyCommonPreset(
...
     // Typescript implements the stage 1 version of decorators, which is the
     // "legacy" version. When decorators are used in legacy mode,
-    // proposal-class-properties must be used in loose mode
-    // see https://babeljs.io/docs/en/babel-plugin-proposal-decorators#note-compatibility-with-babel-plugin-proposal-class-properties
-    typescript && [
+    legacyDecorators && [
       require.resolve('@babel/plugin-proposal-decorators'),
       {legacy: true},
     ],
-    // Enable loose mode to use assignment instead of defineProperty when typescript is enabled
```
dahukish, line 73: These are all now part of the `preset-env` so we don't have to add them again.

packages/babel-preset/index.js
```diff
@@ -64,41 +65,26 @@ module.exports = function shopifyCommonPreset(
     // proposal-decorators must go before proposal-class-properties.
     // Typescript implements the stage 1 version of decorators, which is the
     // "legacy" version. When decorators are used in legacy mode,
```
atesgoral, line 67: This sentence got chopped. Did you mean to remove it entirely?

packages/babel-preset/index.js
```diff
@@ -114,5 +100,17 @@ module.exports = function shopifyCommonPreset(
       require.resolve('@babel/plugin-transform-react-constant-elements'),
   ].filter(Boolean);
 
-  return {presets, plugins};
+  // proposal-class-properties, plugin-proposal-private-methods must be used in loose mode (this is now handled by these assumptions)
```
dahukish, line 103: We can now use assumptions for our legacy decorator plugins. Updated the comments to reflect this.

packages/babel-preset/package.json
```diff
@@ -35,5 +35,8 @@
     "@babel/runtime": "^7.14.6",
     "babel-plugin-react-test-id": "^1.0.2",
     "babel-plugin-transform-inline-environment-variables": "^0.4.3"
+  },
+  "devDependencies": {
+    "@babel/cli": "^7.14.5"
```
dahukish, line 40: need babel cli for testing in this repo.

packages/babel-preset/index.js
vsumner: Why? This is needed if typescript is enabled. Now anywhere typescript is true then legacyDecorators will have to be set?

vsumner:
This will need a beta release test that also confirms shrink-ray and ssr still working.

packages/babel-preset/index.js
dahukish: That's fair. I think @BPScott was thinking this could be more declarative. But if this is only needed for typescript then why break the api.

packages/babel-preset/tests/babel-preset.test.ts
```diff
@@ -0,0 +1,101 @@
...
+import {readFileSync, writeFileSync} from 'fs';
+import {resolve as resolvePath} from 'path';
+
+import shopifyCommonPreset from '../index';
+
+describe('babel-preset-e2e-test', () => {
+  describe('legacy decorator tests', () => {
+    it('runs babel successfully when typescript is true', async () => {
+      const baseConfig = configFactory({typescript: true});
+      writeFileSync(
```
BPScott, line 11: Does thi require writing config to disk and calling the cli? I think we'd be able to do all this in memory using [babel.transform](https://babeljs.io/docs/en/babel-core/#transform)

```js

import { transform } from "@babel/core";
import shopifyPreset from '../index';

// ...

const babelOptions = {
  presets: [shopifyPreset, {typescript: true}]}
};

const output = transform(`SOME CODE HERE`, babelOptions);
const expectedOutput = `BLAH`;

expect(output).toBe(expectedOutput);
```

BPScott: Throwing in a [`configFile: false`](https://babeljs.io/docs/en/options#configfile) to stop it using the default config file would be useful too

BPScott (approved):
We paired and rejigged the tests, and they're really nice

PR #284, opened 2021-08-19
Re-add babel-plugin-react-test-id plugin

BPScott:
## Description

Strip out testID props in non-test environments.

[This was present in v23](https://unpkg.com/browse/@shopify/babel-preset@23.6.2/react.js#L23) but it seems to have been removed when we consolidated all the files into one in v24. I suspect this was accidental as the package is still listed as a dependency in package.json.

While we've been recommending against using testID props for a long time there's still plenty of usage of it web (love them long tails) so it's probably worth keeping this around.

## Type of change

-  babel-preset Patch: Bug (non-breaking change which fixes an issue)


dahukish (approved):
LGTM apart from any context @vsumner might have.

vsumner (approved):
LGTM. It wasn't intentionally removed.

PR #296, opened 2021-10-22
Update ecmaVersion to 'latest' in esnext.js in eslint-plugin to ensure it's esnext.

iamyoki:
## Description
This PR updates a simple config in esnext.js in eslint-plugin, it prevents some potential problems.
It updates
```js
  parserOptions: {
    ecmaVersion: '2018',
    sourceType: 'module',
  }
```
to
```js
  parserOptions: {
    ecmaVersion: 'latest',
    sourceType: 'module',
  }
```
so to ensure it always be esnext latest version.

Here is the official docs mentioned: https://eslint.org/docs/user-guide/configuring/language-options#specifying-parser-options.

Also updates env `es6` to `es2021`.

## Type of change

- [x]  Patch: Bug (non-breaking change which fixes an issue)

## Checklist

### DIFF
--- a/packages/eslint-plugin/lib/config/esnext.js
+++ b/packages/eslint-plugin/lib/config/esnext.js
@@ -9,7 +9,7 @@ module.exports = {
   },
 
   parserOptions: {
-    ecmaVersion: 'latest',
+    ecmaVersion: '2021',
     sourceType: 'module',
   },
 


lucabezerra:
@iamyoki I think you added it twice by mistake, heh :)

packages/eslint-plugin/lib/config/esnext.js
```diff
@@ -5,11 +5,11 @@ module.exports = {
   parser: 'babel-eslint',
 
   env: {
-    es6: true,
+    es2021: true,
   },
 
   parserOptions: {
-    ecmaVersion: 2018,
+    ecmaVersion: 'latest',
```
BPScott, line 12:
```suggestion
    ecmaVersion: '2021',
```

`latest` doesn't seem to be supported in the parsers that we currently use. Let's keep this as `2021`, and look at using `latest` once we're using parsers that support it.

- We use `@typescript-eslint/parser@^4.28.2` however support for ecmaVersion 'latest' was added in [v4.32.0](https://github.com/typescript-eslint/typescript-eslint/releases/tag/v4.32.0)
- We use the long-deprecated `babel-eslint` however  support for ecmaVersion 'latest' was added in it's successor `@babel/eslint-parser` in [v7.15.4](https://github.com/babel/babel/pull/13638)

BPScott (changes requested):
`latest` isn't supported in our the parsers versions that we currently require.

Let's have use `ecmaVersion: 2021` to get us onto the latest version in this PR

Then if you're feeling a brave, open new PRs for replacing `babel-eslint` with `@babel/eslint-parser`, and updating `@typescript-eslint/parser` to the latest version in the v4 branch (we'll want to go to v5 at some point but perhaps we should do that as part of the plugin updates required to support eslint v8).

BPScott (approved):

PR #301, opened 2021-11-18
Babel 7.16 support

vsumner:
## Description
Update the babel-preset to support 7.16 and webpack 4.

Fixes: https://github.com/Shopify/sewing-kit/issues/2875

## Type of change

## Checklist


packages/babel-preset/index.js
```diff
@@ -95,6 +95,16 @@ module.exports = function shopifyCommonPreset(
         ]
       : []),
 
+    // when this is still webpack4 and typescript is not enabled
+    // the following methods must be enabled, this is because
+    // webpack5 is still on acron 6, not required for v5 of webpack
```
GoodForOneFare, line 100:
```suggestion
    // webpack5 is still on acorn 6, not required for v5 of webpack
```

yarn.lock
```diff
@@ -49,6 +61,27 @@
     semver "^6.3.0"
     source-map "^0.5.0"
 
+"@babel/core@^7.16.0":
```
GoodForOneFare, line 64: Can we get rid of the above `@babel/core@7.15.0`?

GoodForOneFare (approved):
👍 Usual concerns about duplicated babel stuff, but the change looks good.

PR #298, opened 2021-11-01
Add Node.js 16 support and update stylelint-plugin deps

atesgoral:
## Description

- Bump the Node.js for local development to 16.3.0.
- Add an `engines` prop to package.json
- Update stylelint-plugin deps:

#### [stylelint-config-prettier 8.0.1 -> 9.0.3](https://github.com/prettier/stylelint-config-prettier/releases)

v9.0.0

- Updates rules
- Adds SCSS rule resets

#### [stylelint-order 4.0.0 -> 5.0.0](https://github.com/hudochenkov/stylelint-order/blob/master/CHANGELOG.md)

5.0.0

- Breaking change: Dropped Node.js 8 support. Node.js 12 or greater is now required.
- Breaking change: Dropped support for Stylelint 13 and earlier.
- Added support for Stylelint 14.

#### [stylelint-scss 3.19.0 -> 4.0.0](https://github.com/stylelint-scss/stylelint-scss/blob/master/CHANGELOG.md)

4.0.0

- Removed: stylelint@13 support.
- Added: stylelint@14 support.

#### [stylelint 13.12.0 -> 14.0.1](https://github.com/stylelint/stylelint/blob/main/CHANGELOG.md)

14.0.0

[Migration guide](https://github.com/stylelint/stylelint/blob/main/docs/migration-guide/to-14.md)

- Removed: Node.js 10 support
- Removed: syntax option
- Removed: configOverrides option
- Removed: rules deprecated in 13.7.0
- Removed: function-calc-no-invalid rule
- Changed: ignoreFiles to be extendable

## Type of change

- [x] `stylelint-plugin` Major: Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

- [x] I have added a changelog entry, prefixed by the type of change noted above (Documentation fix and Test update does not need a changelog as we do not publish a new version for these changes)

PR #302, opened 2021-11-26
Update CI to tests multiple node versions, use node 16 locally

BPScott:
- Run CI against node 12 / 14 / 16. Following a similar pattern to quilt and loom
- Use node 16.13.0 (LTS version) locally

We attempted part of this previously, but bundled it with a bunch of stylelint updates that needed to be reverted in #298

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,22 +2,27 @@ name: CI
 
 on:
   pull_request:
-    branches:
-      - '**'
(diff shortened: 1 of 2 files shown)


atesgoral (approved):

BPScott:
Updated the branch protection rules so now they require the three new test checks (`Test (12.14)` etc) instead of the old singular "Test"

PR #306, opened 2021-11-30
Refactor stylelint-config tests

BPScott:
## Description

Tests within stylelint-config could do with some love.

- Leverage `jest-preset-stylelint` for testing individual rules
- Create a set of e2e tests that execute stylelint and check its output
  for testing rule configuration
- Create an `rules/index.js` file for exporting all custom rules.

This sets the stage for updating stylelint to v14 as its removal of built-in support for scss shall require a bit of e2e testing

## Type of change

N/A - this is test stuff only


packages/stylelint-plugin/rules/index.js
```diff
@@ -0,0 +1 @@
+module.exports = [require('./content-no-strings')];
```
BPScott, line 1: This allows us to add multiple rules to this array and have them all be registered

packages/stylelint-plugin/tests/config-e2e.test.js
```diff
@@ -0,0 +1,41 @@
+const {spawnSync} = require('child_process');
```
BPScott, line 1: Copied from my learnings while wrangling e2e tests in https://github.com/prettier/stylelint-prettier

BPScott:
> Probably needs the change in your other PR bumping node 12.

Yepyep, fixing up some dedupes you mentioned then shall merge that PR and rebase this

PR #307, opened 2021-12-01
Support stylelint v14

BPScott:
## Description

Updates @shopify/styelint-plugin to support stylelint v8.

- Increase peer dependency on stylelint from `>=13.12.0` to `>=14.1.0`
- Add overrides for scss files to configure the custom syntax
- Add e2e test for scss files that demonstrate syntax that would error if
  scss parsers / scss rules were not configured
- Update plugins we depend upon

| Package                     | Old version | New version |
| --------------------------- | ----------- | ----------- |
| `stylelint-config-prettier` | `^8.0.1"`   | `^9.0.3`    |
| `stylelint-order`           | `^4.0.0`    | `^5.0.0`    |
| `stylelint-prettier`        | `^1.2.0`    | `^2.0.0`    |
| `stylelint-scss`            | `^3.19.0`   | `^4.0.0`    |

## Type of change

- stylelint-plugin Major: Bug (non-breaking change which fixes an issue)

### DIFF
--- a/packages/stylelint-plugin/tests/fixtures/scss.invalid.scss
+++ b/packages/stylelint-plugin/tests/fixtures/scss.invalid.scss
@@ -1,3 +1,6 @@
+// CSS nesting trigers a parse error if we're not using a customSyntax
+// selector-max-class/selector-max-combinators are configured to disallow 3 or
+// more levels of nesting
 .n1 {
   .n2 {
     .n3 {
@@ -6,11 +9,15 @@
   }
 }
 
+// scss/dimension-no-non-numeric-values is configured to disallow creating
+// dimention values using string concatenation
 .foo {
   $size: 10;
   padding: #{$size}px;
 }
 
+// scss/selector-no-union-class-name is configured to disallow using & for
+// constructing class names
 .bar {
   &-baz {
     padding: 8px;
--- a/packages/stylelint-plugin/tests/fixtures/scss.invalid.scss
+++ b/packages/stylelint-plugin/tests/fixtures/scss.invalid.scss
@@ -1,4 +1,4 @@
-// CSS nesting trigers a parse error if we're not using a customSyntax
+// CSS nesting triggers a parse error if we're not using a customSyntax
 // selector-max-class/selector-max-combinators are configured to disallow 3 or
 // more levels of nesting
 .n1 {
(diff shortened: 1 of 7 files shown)


packages/stylelint-plugin/tests/fixtures/scss.invalid.scss
```diff
@@ -0,0 +1,18 @@
+.n1 {
```
atesgoral, line 1: Could you add comments to each invalid case to point out what's invalid about it?

packages/stylelint-plugin/tests/fixtures/scss.invalid.scss
```diff
@@ -6,11 +9,15 @@
   }
 }
 
+// scss/dimension-no-non-numeric-values is configured to disallow creating
+// dimention values using string concatenation
```
atesgoral, line 13:
```suggestion
// dimension values using string concatenation
```

packages/stylelint-plugin/tests/fixtures/scss.invalid.scss
```diff
@@ -1,3 +1,6 @@
+// CSS nesting trigers a parse error if we're not using a customSyntax
```
atesgoral, line 1:
```suggestion
// CSS nesting triggers a parse error if we're not using a customSyntax
```

packages/stylelint-plugin/tests/fixtures/scss.valid.scss
```diff
@@ -1,3 +1,4 @@
+// CSS nesting trigers a parse error if we're not using a customSyntax
```
atesgoral, line 1: This is valid, do we need a comment?

packages/stylelint-plugin/tests/fixtures/scss.invalid.scss
BPScott: Done!

packages/stylelint-plugin/tests/fixtures/scss.valid.scss
BPScott: I think it fulfills the same purpose as the comments in invalid.scss - it says what's bit of code is important for this test output to pass.

- In invalid.scss tests will fail because expected stylelint output was not present if you remove the scss rules from the config 
- In valid.scss tests will fail because we expected stylelint output to be empty if you remove the customSyntax from the config

PR #308, opened 2021-12-01
Replace usage of merge with object spreading

BPScott:
## Description

Prior to this PR `eslint-plugin` and `stylelint-plugin` depended upon `merge`. That dependency is unneeded and can be replaced with using object spread syntax.

This PR replaces usage of `merge` with object spread syntax.

## To test

- Create a test.js file in the root of the repo that dumps out the eslint and stylelint configs:
  ```js
    console.log(
      JSON.stringify(require('./packages/eslint-plugin'), null, 2),
      '\n\n\n\n',
      JSON.stringify(require('./packages/stylelint-plugin'), null, 2),
    );
  ```
- Run that test in the main branch and save the output `node test.js > main.txt`
- Run that test in this branch branch and save the output `node test.js > branch.txt`
- Run a diff between those two files and note that they are identical

## Type of change

- [ ] eslint-plugin Patch: Bug (non-breaking change which fixes an issue)
- [ ] stylelint-plugin Patch: Bug (non-breaking change which fixes an issue)


atesgoral (approved):
Changed look sane. Didn't tophat.

I trust ya ;)

PR #309, opened 2021-12-01
Stylelint - flatten configs

BPScott:
## Description

Stylelint config settings are hard to find when they're spread across multiple files. It'd be easier to navigate if all we had files that configured rules by "here is everything provided by stylelint core" and then multiple per-plugin rules.

This PR consolidates several files in packages/stylelint-plugin/config/* into a single `core.js` file that contains all the rules that are defined in stylelint's core.

## To test

- Create a test.js file in the root of the repo that dumps out the eslint and stylelint configs:
  ```js
    console.log(
      JSON.stringify(require('./packages/eslint-plugin'), null, 2),
      '\n\n\n\n',
      JSON.stringify(require('./packages/stylelint-plugin'), null, 2),
    );
  ```
- Run that test in the main branch and save the output `node test.js > main.txt`
- Run that test in this branch branch and save the output `node test.js > branch.txt`
- Run a diff between those two files and note the ordering of some items has changed but they're all present


atesgoral (approved):
Amazing!

PR #312, opened 2021-12-08
Update svgo to 2.4 and remove extendDefault

vsumner:
## Description
In `svgo` `2.4.0` `extendDefaultPlugins` was deprecated. This PR removes the use of `extendDefaultPlugins` and requires `2.4.0` of `svgo` at minimum.

## Type of change

- [x] `@shopify/images` Minor: New feature (non-breaking change which adds functionality)

## Checklist

### DIFF
--- a/packages/images/optimize.js
+++ b/packages/images/optimize.js
@@ -1,10 +1,15 @@
-const {extendDefaultPlugins} = require('svgo');
-
 function svgOptions() {
   return {
-    plugins: extendDefaultPlugins([{name: 'removeTitle'}]).filter(
-      (plugin) => plugin.name !== 'removeViewBox',
-    ),
+    plugins: [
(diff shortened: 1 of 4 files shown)


packages/images/optimize.js
```diff
@@ -1,10 +1,15 @@
-const {extendDefaultPlugins} = require('svgo');
-
 function svgOptions() {
   return {
-    plugins: extendDefaultPlugins([{name: 'removeTitle'}]).filter(
-      (plugin) => plugin.name !== 'removeViewBox',
-    ),
+    plugins: [
+      {
+        name: 'preset-default',
```
vsumner, line 5: `removeTitle` is enabled by default now.

caropinzonsilva (approved):

PR #313, opened 2021-12-17
Babel - move options defaults out of default values

BPScott:
## Description

Previously as soon as you specified any value for `reactOptions` or `transformRuntimeOptions` it would throw any of our default values.

For instances if you specified `{reactOptions: {runtime:'automatic'}}` then it would throw away our modified default value for `useBuiltIns`. Specifying `{}` would throw away all our defaults.

By moving the defaults down to where the various presets get called users who specify options won't throw away / have to recreate all our defaults.

## Type of change

- babel-preset: Breaking

## Context

I'm considering a future where we start wanting folks to enable the automatic babel runtime.

Prior to this PR doing `['@shopify/babel-preset, {reactOptions: {runtime: 'automatic'}}]` would result in `useBuiltIns` being the babel default of false, instead of our value of true. Forcing consumers to do `{reactOptions: {runtime: 'automatic', useBuiltins: true}}` just to keep our default desired behaviour seems real clunky and undexpected.

### DIFF
--- a/packages/babel-preset/index.js
+++ b/packages/babel-preset/index.js
@@ -8,24 +8,9 @@ module.exports = function shopifyCommonPreset(
     typescript = false,
     typescriptOptions = {},
     transformRuntime = false,
-    transformRuntimeOptions = {
-      corejs: false,
-      helpers: true,
-      // By default, babel assumes babel/runtime version 7.0.0-beta.0,
-      // explicitly resolving to match the provided helper functions.
-      // https://github.com/babel/babel/issues/10261
-      version: require('@babel/runtime/package.json').version,
-      regenerator: true,
-      // This allows users to run transform-runtime broadly across a whole project.
-      // By default, transform-runtime imports from @babel/runtime/foo directly, but
-      // that only works if @babel/runtime is in the node_modules of the file that is being compiled.
-      absoluteRuntime: false,
-    },
+    transformRuntimeOptions = {},
     react = false,
-    reactOptions = {
-      // Will use the native built-in instead of trying to polyfill behavior for any plugins that require one.
-      useBuiltIns: true,
-    },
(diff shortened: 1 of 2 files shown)


packages/babel-preset/index.js
```diff
@@ -110,8 +97,18 @@ module.exports = function shopifyCommonPreset(
     transformRuntime && [
       '@babel/plugin-transform-runtime',
       {
-        ...transformRuntimeOptions,
+        corejs: false,
+        helpers: true,
+        // By default, babel assumes babel/runtime version 7.0.0-beta.0,
+        // explicitly resolving to match the provided helper functions.
+        // https://github.com/babel/babel/issues/10261
         version: require('@babel/runtime/package.json').version,
```
BPScott, line 105: Previously version came after `...transformRuntimeOptions` which meant that it was always set. Now the user can override it. I can't ever thing of a reason why they'd want to do that but it feels more flexible than forcing it.

BPScott:
This was originally spurred by future plans of "if we get consumers to set `reactOptions: {runtime: 'automatic'}` then it'll overwrite the `useBuiltIns: true` that is currently set, and thus we'd need to make people do `reactOptions: {runtime: 'automatic', useBuiltIns: true}`. However reading https://babeljs.io/docs/en/babel-preset-react/#react-classic-runtime it suggests that `useBuiltIns` is only a valid option when using the classic runtime. It has no effect when using the automatic runtime. So it's totally fine for users to do `reactOptions: {runtime: 'automatic'}` and blow away the useBuiltIns option.

In fact it'll make things harder in the future as once Babel 8 eventually lands this usage of `useBuiltIns` will throw an error and with this PR it will be impossible to remove that option.

With that in mind there's no value in merging this PR now.

PR #315, opened 2022-01-10
[Unified Admin] Restrict /admin in Route path literals and templates

ttemrahcelik:
## Description

We are on our way to remove the path prefix `/admin` as part of Unified Admin project.
The prefix is being extracted into a global state/variable to be shared across web as much as we can.
In the meantime, we draw the line with this rule so we don't keep adding routes with the hardcoded prefix.

## Type of change

- [x]  Major: Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

### DIFF
--- a/packages/eslint-plugin/CHANGELOG.md
+++ b/packages/eslint-plugin/CHANGELOG.md
@@ -7,6 +7,12 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 <!-- ## Unreleased -->
 
+## 42.0.0 - 2022-01-17
+
+### Breaking Change
+
+- Restrict /admin in Route path literals and templates [[#315](https://github.com/Shopify/web-configs/pull/315)]
+
 ## 41.0.1 - 2021-12-02
 
 ### Changed
--- a/packages/eslint-plugin/README.md
+++ b/packages/eslint-plugin/README.md
@@ -132,6 +132,7 @@ This plugin provides the following custom rules, which are included as appropria
 - [images-no-direct-imports](docs/rules/images-no-direct-imports.md): Prevent images from being directly imported.
 - [jest/no-snapshots](docs/rules/jest/no-snapshots.md): Disallows jest snapshots.
 - [jsx-no-complex-expressions](docs/rules/jsx-no-complex-expressions.md): Disallow complex expressions embedded in in JSX.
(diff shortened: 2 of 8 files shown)


BPScott:
This stuff is all seems very web-specific. Does this have use outside of the web repository?

I wonder if we should be putting web-specific rules directly in the web repository

ttemrahcelik:
> This stuff is all seems very web-specific. Does this have use outside of the web repository?
> 
> I wonder if we should be putting web-specific rules directly in the web repository

The stuff I'm adding is only for web.

You mean moving this whole project to web, or only what I'm adding?

And may I ask where we add these rules to web?

Shipping from 1 repo is much easier, but if we're talking about merging 2 repos, even though it doesn't look too difficult, I'm not sure how much time I can spare for it.

ciuliano-shopify (approved):
LGTM!
Seems odd to me that this rule lives outside of the Web repository though.
I sent you a DM on Slack about the approach - I think we might not need a lint rule if we handle this differently but IDK for sure.

BPScott:
> You mean moving this whole project to web, or only what I'm adding?

I meant define this one lint rule into web, as this lint rule is only applicable in this the web repository. Putting something that only applies to a single repository into a config that is applicable to multiple projects feels strange. Web-specific things should go in web.

> And may I ask where we add these rules to web?

We don't currently have a folder for this sort of thing. But with [impending work](https://github.com/Shopify/web/issues/55393) we'll need somewhere to put build/test/lint tooling config. Perhaps we create a new top level folder for storing general build config, then this could live in `build-config/eslint/rules/jsx-no-hardcoded-admin-prefix`?


> Shipping from 1 repo is much easier, but if we're talking about merging 2 repos, even though it doesn't look too difficult, I'm not sure how much time I can spare for it.

You wouldn't need to port across all the existing content in the eslint-plugin package. Just add this rule into web, instead of here.

BPScott (changes requested):
Marking this as "request changes" while we work out if this could live directly in the web repository instead of putting a web-specific rule into pan organisation config.

ttemrahcelik:
> You wouldn't need to port across all the existing content in the eslint-plugin package. Just add this rule into web, instead of here.

ok i'm all for this. just wanted to make sure if this whole `web-configs` project is web specific or not. 🍻

BPScott:
> just wanted to make sure if this whole web-configs project is web specific or not. 

Stuff in this repo is org-wide :) The eternally confusing question "does web mean web as a discipline within shopify, or just the admin web codebase" strikes again.

ttemrahcelik:
Closing this in favour of adding the rule as part of: https://github.com/Shopify/web/pull/55083

PR #319, opened 2022-02-01
Remove lines around comment rule in prettier eslint-plugins

royleejr:
## Description

As said in [this PR](https://github.com/Shopify/web/pull/56144). We have discussed and decided that the rule `lines-around-comment` should no longer be enforced upon us because of known typescript issues such as [typescript-eslint/typescript-eslint#1150](https://github.com/typescript-eslint/typescript-eslint/issues/1150). We believe that we should stay consistent with this rule all across `shopify` and should remove this rule from the `eslint-plugins`.

## Type of change

- [x] eslint-plugin Minor: New feature (non-breaking change which adds functionality)

## Checklist

### DIFF
--- a/packages/eslint-plugin/CHANGELOG.md
+++ b/packages/eslint-plugin/CHANGELOG.md
@@ -5,7 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-<!-- ## Unreleased -->
+## Unreleased
+
+### Changed
+
+- Removed `lines-around-comment` rule, let prettier handle formatting of comments. [[#319](https://github.com/Shopify/web-configs/pull/319)]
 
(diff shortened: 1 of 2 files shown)


packages/eslint-plugin/CHANGELOG.md
```diff
@@ -6,6 +6,11 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 <!-- ## Unreleased -->
+## Unreleased
+
+### Changed
+
+- Removed `lines-around-comment` rule in `eslint-plugin`. [[#319](https://github.com/Shopify/web-configs/pull/319)]
```
BPScott, line 13:
```suggestion
- Disabled `lines-around-comment` rule, let prettier handle formatting of comments. [[#319](https://github.com/Shopify/web-configs/pull/319)]
```

"in `eslint-plugin" isn't really needed as this is the CHANGELOG for the eslint-plugin package

BPScott (approved):

PR #326, opened 2022-03-10
upgrade typescript and ts-eslint

m4thieulavoie:
## Description

This PR upgrades a few packages like `typescript` and `@typescript-eslint/`

This should remove the 

```
=============

WARNING: You are currently running a version of TypeScript which is not officially supported by @typescript-eslint/typescript-estree.

You may find that it works just fine, or you may not.

SUPPORTED TYPESCRIPT VERSIONS: >=3.3.1 <4.6.0

YOUR TYPESCRIPT VERSION: 4.6.2

Please only submit bug reports when using the officially supported version.

=============
```

warnings

## Type of change

- [x] Patch

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
   "devDependencies": {
     "@shopify/tophat": "^0.0.4",
     "@types/jest": "^26.0.24",
-    "eslint": "^8.3.0",
+    "eslint": "^8.10.0",
     "fs-extra": "^9.0.0",
     "glob": "^7.1.6",
     "jest": "^27.0.6",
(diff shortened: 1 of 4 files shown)


BPScott:
While keeping up to date is nice, this isn't required - the current version range of `^5.4.0` allows for `5.14.0` and other minor versions beyond it to be installed. Unless we need folks to update because of we want to take advantage of some new feature then I'm not super inclined to accept PRs that bump versions - it's inventing busywork  to merge and release this, and for consumers to update.

PR #329, opened 2022-03-29
Add `q` as allowed identifier in eslint

ianks:
## Description

`q` is a common identifier for a search query param, and it used in many apps across Shopify. 

Previously, we had to do this:

```js
// eslint-disable-next-line
const { data, loading, error } = useSearchPackages({ q: "foo" })
```

This patch makes it less painful to use `q` in Javascript-land.

## Type of change

- [ ] @shopify/eslint-plugin Patch: Bug (non-breaking change which fixes an issue)
- [x] @shopify/eslint-plugin Minor: New feature (non-breaking change which adds functionality)
- [ ] @shopify/eslint-plugin Major: Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

### DIFF
--- a/packages/eslint-plugin/CHANGELOG.md
+++ b/packages/eslint-plugin/CHANGELOG.md
@@ -7,10 +7,6 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 <!-- ## Unreleased -->
 
-### Changed
-
-- Allow usage of `q` variable in `id-length` rule [[#329](https://github.com/Shopify/web-configs/pull/329)]
-
 ## 41.2.0 - 2022-03-07
 
 ### Changed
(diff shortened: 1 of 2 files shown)


ianks:
Couldn't figure out the changelog test... but here's what I was going to add if it helps:

```
### Changed

- Allow usage of `q` variable in `id-length` rule [[#329](https://github.com/Shopify/web-configs/pull/329)]
```

BPScott:
@ianks Regarding the changelog - uncomment the `## Unreleased` header, in addition to adding the content in your prior post.

BPScott (approved):
I took the liberty of pushing up the correction to the chnagelog.

Thanks!

BPScott:
Published as 41.2.1!

PR #338, opened 2022-06-27
Flatten eslint configs, removing config/rules

BPScott:
## Description

The config definitions in `lib/configs`, import contents from `lib/configs/rules`. This extra layer of indirection adds no value - nothing is ever reused.

This PR flattens this indirection, replacing `require` calls in `lib/configs/*.js` with the contents of the files being required.

This has no affect on the rulesets used, it's just tidying.

## To test

Compare the evaluated output of the `lib/config/*.js` files on main, and in this branch and note that they are identical.

Copy the following script into the root of the repo as `test.js`:

```js
const glob = require('glob');

const fileEval = glob
  .sync('./packages/eslint-plugin/lib/config/*.js')
  .map((filename) => {
    return [filename, JSON.stringify(require(filename), validatingReplacer, 2)];
  });

const formattedResult = fileEval
  .map(
    ([filename, evaluatedContent]) =>
      `## Filename: ${filename}\n\n\`\`\`\n${evaluatedContent}\n\`\`\``,
  )
  .join('\n\n');

console.log(formattedResult);

//
// UTILS
//

function validatingReplacer(key, value) {
  if (
    ['number', 'string', 'boolean'].includes(typeof value) ||
    Array.isArray(value) ||
    isPlainObject(value) ||
    value === null
  ) {
    return value;
  }

  console.log({key, value});
  throw new Error(`Found an unknown value for key ${key}!`);
}

function isPlainObject(value) {
  if (typeof value !== 'object' || value === null) {
    return false;
  }

  const prototype = Object.getPrototypeOf(value);
  return (
    (prototype === null ||
      prototype === Object.prototype ||
      Object.getPrototypeOf(prototype) === null) &&
    !(Symbol.toStringTag in value) &&
    !(Symbol.iterator in value)
  );
}

```

- Run the script on the main branch, run the script and save the output to a file `node test.js > main-config-content.md`
- Check out this branch, run the script and save the output to a file: `node test.js > branch-config-content.md`
- Diff the output of those two files and note that they are identical `diff -u main-config-content.md branch-config-content.md`

Note that modifying some file in lib/config and doing the above again shall result in some diff output to prove that changes are noticed.

### DIFF
--- a/packages/eslint-plugin/lib/config/jest.js
+++ b/packages/eslint-plugin/lib/config/jest.js
@@ -88,7 +88,19 @@ module.exports = {
     // Prevents `expect` statements outside of a `test` or `it` block
     'jest/no-standalone-expect': 'error',
     // Enforce titles don't duplicate the test function name or start with a space
-    'jest/valid-title': 'error',
+    'jest/valid-title': [
+      'error',
+      {
+        disallowedWords: [
+          'correct',
+          'appropriate',
+          'properly',
+          'should',
+          'every',
+          'descriptive',
+        ],
+      },
+    ],
     // Suggest to have all hooks at top-level before tests
     'jest/prefer-hooks-on-top': 'error',
     // Require top-level describe block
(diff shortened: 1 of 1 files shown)


packages/eslint-plugin/index.js
```diff
@@ -40,15 +40,15 @@ module.exports = {
...
-    react: require('./lib/config/react'),
     typescript: require('./lib/config/typescript'),
-    'typescript-type-checking': require('./lib/config/typescript-type-checking'),
 
     // Augmenting configs - When extending, these go after the core config
     graphql: require('./lib/config/graphql'),
     jest: require('./lib/config/jest'),
     node: require('./lib/config/node'),
     polaris: require('./lib/config/polaris'),
+    react: require('./lib/config/react'),
```
BPScott, line 50: React and typescript-type-checking are both augment configs, not core configs, per the README.

PR #347, opened 2022-09-14
[babel-preset] Allow customizing Babel assumptions and set loose~ish defaults

developit:
This adds support for customizing the [`assumptions` configuration](https://babeljs.io/docs/en/assumptions) passed to `@babel/preset-env`, which is a granular replacement for Babel's previous "loose mode" functionality.

I've also included a set of default assumptions, with comments above each explaining what they do. Some of them are clear-cut "we would never rely on that" assumptions (`document.all`), others I've tried to infer from the TS+JS I've scanned through via `cs.github.com`.

The three I'm most concerned about would be `objectRestNoSymbols`, `setSpreadProperties` and `superIsCallableConstructor`. These all have good runtime performance payoff if we can land them by default - the normal "strict" helpers do extra iteration that the "loose" helpers can skip. If anyone believes Shopify might currently be relying on this functionality, we'd want to remove them.

### DIFF
--- a/.changeset/chatty-poems-clap.md
+++ b/.changeset/chatty-poems-clap.md
@@ -1,0 +1,5 @@
+---
+'@shopify/babel-preset': minor
+---
+
+Allow customizing Babel assumptions, and adopt some reasonably loose defaults rather than the normal strict defaults.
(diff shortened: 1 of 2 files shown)


developit:
Good call @BPScott. Thanks for the merge!

PR #350, opened 2022-12-05
Update eslint-plugin-import

marvinhagemeister:
## Description

This addresses a performance issue caused by an issue in the underlying `resolve` dependency that is used under the hood (see https://github.com/browserify/resolve/commit/1382486e2c1063393f6e6856f2e68330e67a7bc5). The issue is that the overhead of capturing stack traces is very significant in our setup. It's probably a symptom of something else going on, but just updating this dependency leads to an immediate win in performance.

Times when linting:

| Description | Time |
|---|--:|
| Before | 2:26 min |
| After (this PR) | **1:51 min** |

### DIFF
--- a/.changeset/eleven-seals-move.md
+++ b/.changeset/eleven-seals-move.md
@@ -1,0 +1,5 @@
+---
+'@shopify/eslint-plugin': patch
+---
+
+Update the `eslint-plugin-import` to the latest version which contains significant performance improvements for module resolution.
--- a/packages/eslint-plugin/package.json
+++ b/packages/eslint-plugin/package.json
@@ -35,7 +35,7 @@
     "eslint-config-prettier": "^8.3.0",
     "eslint-module-utils": "^2.7.1",
     "eslint-plugin-eslint-comments": "^3.2.0",
-    "eslint-plugin-import": "^2.25.3",
+    "eslint-plugin-import": "^2.26.0",
     "eslint-plugin-jest": "^25.3.0",
     "eslint-plugin-jest-formatting": "^3.1.0",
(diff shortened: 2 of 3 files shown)


simontaisne (approved):
🥇

notjosh:
@marvinhagemeister Heya! Small followup, because we bumped into it on Shop also: `yarn.lock` (here, and likely your project) still has references to `resolve < 1.22.1` (and `resolve <= 2.0.0-next.4`, used only by `eslint-plugin-react`), so you can likely get a similar boost in some other tooling - we found Jest performance increased noticeably (via `jest-resolve` to `resolve`) when we bumped the version.

Two things you'll need to do on your project:

1. (for `resolve@1.x`) run `npx yarn-deduplicate --packages resolve`, which should bump your `yarn.lock` to force the `1.x` references to the newer one. Easy.
2. (for `resolve@2.x`) we'll temporarily use `resolutions` to bump the subdependency (afaik there's still no better way to do this ([ref](https://github.com/yarnpkg/yarn/issues/4986))
  1. add `"eslint-plugin-react/resolve": "2.0.0-next.4"` to `package.json` resolutions
  2. run `yarn --check-files` to install the new version
  3. remove the custom resolution from `package.json`, we don't need it now that `yarn.lock` is updated
  4. run `npx yarn-deduplicate --packages resolve` to remove any `yarn.lock` references to `2.0.0-next.3`
  5. verify: at this point, only `yarn.lock` should have a diff

You can verify it's set up correctly by running `yarn why resolve`, and only the newer versions should be installed.

Note: We _can_ do that here on this repo too (there's no reason not to!), but that won't force the updated version in other projects.

Hope that's helpful!

PR #356, opened 2023-01-09
Cache repeated directory lookups for 12% speedup

marvinhagemeister:
## Description

While profiling eslint tasks in various projects, I noticed that the polaris-related linting rules we have, take up a more time than necessary. This is most noticeable when when the project contains lots of files.

Screenshot 2023-01-09 at 19 03 01

The majority of the time in those two rules is spent in `normalizeSource` which traverses directories upwards to find the npm package root. This is perfectly fine if we haven't seen the file before, but unnecessarily expensive when we _did_ resolve it once already. In the projects I tried this PR on, 95% of calls to `normalizeSource` are with a source file that we resolved already.

Therefore adding a simple caching mechanism bypasses the directory traversal and leads to about 12% faster linting times on the project I've tested it on.

### DIFF
--- a/.changeset/violet-laws-provide.md
+++ b/.changeset/violet-laws-provide.md
@@ -1,0 +1,5 @@
+---
+'@shopify/eslint-plugin': patch
+---
+
+Speedup polaris linting rules by about 12% via caching already resolved files.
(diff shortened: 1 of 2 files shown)


BPScott (approved):
Great find!

Merge when you're happy, and I'll cut a release.

PR #358, opened 2023-01-20
update eslint versions

m4thieulavoie:
## Description

This PR bumps the versions of each `eslint` dependencies. The main motivation behind that change is to get `@typescript-eslint` to handle typescript's new `satisfies` keyword


packages/eslint-plugin/package.json
```diff
@@ -25,32 +25,32 @@
...
+    "eslint-plugin-react": "^7.32.0",
+    "eslint-plugin-react-hooks": "^4.6.0",
+    "eslint-plugin-sort-class-members": "^1.16.0",
     "jsx-ast-utils": "^3.2.1",
     "pkg-dir": "^5.0.0",
     "pluralize": "^8.0.0"
   },
   "peerDependencies": {
-    "eslint": "^8.3.0"
+    "eslint": "^8.32.0"
```
BPScott, line 53: Changing the peer dependency on eslint is a breaking change. What package causes this bump and can it be reverted so that we don't force consumers to upgrade more than required?

.changeset/rude-goats-taste.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/eslint-plugin': patch
+---
+
+bump eslint dependencies
```
BPScott, line 5: This is a `major` breaking change thanks to the removal of the the `jest/no-jest-import` rule. Please update the update type.


Please write a more complete changelog entry. Please detail what packages were bumped, and to what version, such as https://github.com/Shopify/web-configs/blob/main/packages/eslint-plugin/CHANGELOG.md#breaking-change. You should also detail any potentially breaking changes - notable the removal of the `jest/no-jest-import` rule.

BPScott (changes requested):
Thanks for the updates. 

A few tidying tasks that should be done prior to merge:

- Deduplicate the yarn.lock file with `npx yarn-dedupe`
- What required the eslint peerDependency bump? Can it be lowered?
- Please write a more complete changeset entry. This is a breaking change as it removes rules - those need to be documented along with the update type being marked as `major`

packages/eslint-plugin/package.json
m4thieulavoie: It [needs](https://github.com/typescript-eslint/typescript-eslint/compare/v5.44.0...v5.45.0#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R82) to be`^8.15.0`. Happy to match that requirement if you judge is better 👍

BPScott:
Closing as stale

PR #360, opened 2023-01-29
Bump browserslist so that the latest NodeJS versions are available

alex-page:
## Description

When getting `@shopify/polaris` ready for v11. I noticed that `browserslist` could not find NodeJS `^v16.17`. This PR updates `browserslist` so that it hopefully bumps the `node-releases` npm package and support.

_After doing some investigation_ I don't think this is blocking our NodeJS upgrade. However it might still be a nice bump.

### DIFF
--- a/.changeset/polite-llamas-confess.md
+++ b/.changeset/polite-llamas-confess.md
@@ -1,0 +1,5 @@
+---
+'@shopify/browserslist-config': minor
+---
+
+Update minumum supported NodeJS versions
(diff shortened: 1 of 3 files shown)


BPScott:
You're changing a devDependency. This has no effect on packages that depend upon `browserslist-config`. Thus it doesn't warrant any changelogs either because the config you've touched is local just to this repository.

As it sounds like you've already discovered, Polaris is free to update the versions of browserslist/node-releases by modifying its own yarn.lock.

I'm going to close this as the change isn't required.

PR #364, opened 2023-02-22
Remove deprecated stylelint rules

alexandcote:
## Description
- `stylelint@^15` deprecated [76 rules](https://stylelint.io/user-guide/rules/#deprecated). 
- Enabling the new [`declaration-property-value-no-unknown`](https://github.com/stylelint/stylelint/blob/main/lib/rules/declaration-property-value-no-unknown/README.md) rule.


BPScott (commented):
If we're dropping stylelint <15 support then we should make sure that all the plugins we use support v15 in their peer dependencies

At this point is it saying "we always expect you to be using prettier" and roll the contents of prettier.js into the index.js?

BPScott (approved):

PR #371, opened 2023-03-09
[eslint plugin] Update `polaris-no-bare-stack-item` and `polaris-prefer-sectioned-prop`

laurkim:
## Description

Part of [Polaris#8185](https://github.com/Shopify/polaris/issues/8185).

The Polaris style guide was updated to group components into categories. This updates links to Polaris components to reflect those changes.

Additionally, `Stack` and `Card` have been deprecated in favor of `LegacyStack` and `LegacyCard`. 
I've updated the `polaris-no-bare-stack-item` and `polaris-prefer-sectioned-prop` rules and tests to use the legacy components instead.

### DIFF
--- a/.changeset/brave-masks-march.md
+++ b/.changeset/brave-masks-march.md
@@ -1,0 +1,5 @@
+---
+'@shopify/eslint-plugin': patch
+---
+
+Updated component URLs for polaris.shopify.com and updated `polaris-no-bare-stack-item` to use `LegacyStack` and `polaris-prefer-sectioned-prop` to use `LegacyCard`'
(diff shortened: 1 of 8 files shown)


laurkim:
@BPScott This was originally opened with the intent just to change the URLs (in a patch change) and then I saw the rules and thought it would be ideal to update them. It was an error on my part to leave this as a patch. I can revert these and re-add the changes in 2 PRs, a patch change to update the URL and a minor change to update the rules so that Card and Stack as well as their legacy components are considered.

BPScott:
No need to revert and redo from scratch over 2 PRs, I'm fine with just one new PR to fix forwards. It's trivial to update the existing changeset file with the new target of minor

PR #379, opened 2023-03-30
[eslint plugin] Add back support for `Card` and `Stack` to polaris rules

laurkim:
## Description

Part of [Polaris#8185](https://github.com/Shopify/polaris/issues/8185).

A previous [PR](https://github.com/Shopify/web-configs/pull/371) updated the `polaris-no-bare-stack-item` to only support `LegacyStack`. Similarly, the `polaris-prefer-sectioned-prop` to only support `LegacyCard`, all in a patch release.

This updates those rules (along with documentation and tests) to support both `Card` and `Stack` along with their legacy counterparts, `LegacyCard` and `LegacyStack` in a minor release. More context [here](https://github.com/Shopify/web-configs/pull/371#issuecomment-1482160392).

### DIFF
--- a/.changeset/brave-masks-march.md
+++ b/.changeset/brave-masks-march.md
@@ -1,5 +1,5 @@
 ---
-'@shopify/eslint-plugin': patch
+'@shopify/eslint-plugin': minor
 ---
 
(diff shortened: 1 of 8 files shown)


BPScott (commented):
Looks good to me, thanks! One tiny typo to be fixed then this can be squash and merged!

PR #397, opened 2023-08-04
upgrade prettier plugins

m4thieulavoie:
## Description

This PR upgrades prettier plugin versions so we can upgrade `web`'s version to ^3


BPScott:
Also if you're in the process of making breaking changes - `@typescript/eslint-plugin` v6 recently got released and we'll need to update to use that if we want to support Typescript 5.2 or later.

BPScott (approved):
Thanks for wrangling the node bumps along the way, thanks to new version of execa that gets pulled in wanting node `^14.18.0 || ^16.14.0 || >=18.0.0`.

Jumping to 18.16 for the in-repo node version makes sense as that's the most recent LTS. We're still running CI against 14/16/18 and we don't force versions through our own engines fields.

PR #420, opened 2024-05-17
Update eslint plugin versions

BPScott:
Major updates:
- Update required node version to `18.20.0`
- Update required eslint version to `^8.56.0`
- Update `@typescript-eslint/eslint-plugin` to `^7.9.0`
- Update `eslint-plugin-jest` to `^28.5.0`
- Update `eslint-config-prettier` to `^9.1.0`

The node and eslint version bumps are a result of updating `@typescript-eslint/eslint-plugin` to v7.

---

Most notably, updating `@typescript-eslint/eslint-plugin` fixes the `WARNING: You are currently running a version of TypeScript which is not officially supported by @typescript-eslint/typescript-estree.` warnings when using Typescript 5.4.0 and above


BPScott:
Uugh, I'm gonna need somebody with adminship to remove "Test (16.x)" as a required commit, as eslint will now no longer run on node 16.

BPScott:
Working pretty well over here: https://github.com/Shopify/checkout-web/pull/34071

Main gotcha is that `jest/no-conditional-in-test` catches a bunch more cases compared to `jest/no-if`, so updating will require a bunch of ignoring those new violations.

PR #426, opened 2024-07-03
Bump stylelint-scss to 6.3.2

kyledurand:


.vscode/settings.json
```diff
@@ -1,6 +1,6 @@
 {
   "editor.codeActionsOnSave": {
-    "source.fixAll.eslint": true
+    "source.fixAll.eslint": "explicit"
```
kyledurand, line 3:
```suggestion
    "source.fixAll.eslint": true
```

BPScott (commented):
Can you also update stylelint-prettier to v5, which adds support for stylelint v16 - https://github.com/prettier/stylelint-prettier/blob/main/CHANGELOG.md

.vscode/settings.json
BPScott: "explict" is desired (though we should stick that in a separate PR), as `source.fixAll` now expect a string rather than a boolean value, and this changing is VSCode updating itself to use the new value: https://code.visualstudio.com/updates/v1_83#_code-actions-on-save-and-auto-save

aaronccasanova (approved):

packages/stylelint-plugin/tests/config-e2e.test.js
```diff
@@ -15,16 +15,23 @@ describe('stylelint-plugin E2E Tests', () => {
     const result = runStylelint('value-keyword-case.*.scss');
 
     const expectedResult = `
+Deprecation warnings:
+ - The "scss/at-import-no-partial-leading-underscore" rule is deprecated.
```
BPScott, line 19: We should fix these dependency warnings by replacing our mention of `scss/at-import-no-partial-leading-underscore` with `scss/load-no-partial-leading-underscore`- https://github.com/Shopify/web-configs/blob/8ff1a906437cdcc6ab33e2eb173371ff769fcc8c/packages/stylelint-plugin/index.js#L479

kyledurand: https://github.com/Shopify/web-configs/pull/428

PR #430, opened 2024-07-04
Remove strip-ansi, use github formatter

kyledurand:
Following what was done here https://github.com/prettier/stylelint-prettier/pull/358


BPScott (approved):
Nice!

I'd be tempted to skip the changeset for this PR as it's an internal facing tests-only change. When folks are reading changelogs as part of updating `@shopify/stylelint-plugin` they don't care that we adjusted out e2e tests and removed a devDependency.

PR #435, opened 2024-09-08
update Stylelint to v16.9.0

romainmenke:
Hi 👋 

We started [monitoring prominent Stylelint packages](https://github.com/stylelint/stylelint-ecosystem-tester) to ensure that we don't accidentally break things when rolling out new features.

We noticed that `@shopify/stylelint-plugin` was failing these tests for two reasons.

------

1) It uses `--formatter=github` to have stable lint output that is easily testable.
This formatter was deprecated and the deprecation warning alone was causing the tests to fail.
I've switched this to `--formatter=json` which should also give stable outcomes.

2) There were also various fixes in Stylelint rules to the reported start and end positions.
This changed the expected outcome of one of the tests.


packages/stylelint-plugin/tests/config-e2e.test.js
```diff
@@ -43,9 +43,13 @@ describe('stylelint-plugin E2E Tests', () => {
...
-  const result = spawnSync(stylelintCmd, ['--formatter=github', pattern], {
-    cwd: path.resolve(__dirname, 'fixtures'),
-  });
+  const result = spawnSync(
+    stylelintCmd,
+    ['--custom-formatter=../../tests/formatter-github.js', pattern],
+    {
+      cwd: path.resolve(__dirname, 'fixtures'),
+    },
+  )
```
romainmenke, line 52:
```suggestion
  );
```

romainmenke: just triggering the CLA check

romainmenke:
Cool! thanks 🙇

PR #438, opened 2024-10-20
Replace deprecated `at-import-partial-extension` stylelint rule

brendo:
## Description

This PR replaces the deprecated stylelint rule with it's replacement as per the warning messages when running stylelint:

```
Deprecation warnings:
 - The "scss/at-import-partial-extension" rule is deprecated.
 - 'at-import-partial-extension has been deprecated, and will be removed in '7.0'. Use 'load-partial-extension' instead. See: https://github.com/stylelint-scss/stylelint-scss/blob/v6.3.0/src/rules/at-import-partial-extension/README.md
```

The error message is helpful, but the link is not. The replacement rule is `load-partial-extension` and the docs for that are [here](https://github.com/stylelint-scss/stylelint-scss/tree/master/src/rules/load-partial-extension).

### DIFF
--- a/.changeset/big-birds-live.md
+++ b/.changeset/big-birds-live.md
@@ -1,0 +1,5 @@
+---
+'@shopify/stylelint-plugin': patch
+---
+
+Replace deprecated `at-import-partial-extension` rule with `load-partial-extension`.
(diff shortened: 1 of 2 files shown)


packages/stylelint-plugin/package.json
```diff
@@ -29,7 +29,7 @@
     "postcss-scss": "^4.0.9",
     "stylelint-order": "^6.0.4",
     "stylelint-prettier": "^5.0.0",
-    "stylelint-scss": "^6.3.2"
+    "stylelint-scss": "^6.8.1"
```
brendo, line 32: The updated rule was introduced in 6.4 (https://github.com/stylelint-scss/stylelint-scss/pull/998)

PR #447, opened 2024-12-14
test snapit fix

BPScott:
Testing snapit fix in https://github.com/Shopify/snapit/pull/24


BPScott:
Blah, the action content reads from main so  need to merge to test.

PR #465, opened 2025-03-10
Add lint rule to disallow contextmenu listeners (right-click)

PhilippeCollin:
## Description

Add an eslint rule to disallow `contextmenu` event listeners. Overriding the default right click menu is generally unexpected and results in worse accessibility. 

At Shopify, we believe that this should almost never been done, except in some full-screen applications, like the Online Store Theme Editor, Figma, Google Docs, etc...

### DIFF
--- a/.changeset/calm-mirrors-fail.md
+++ b/.changeset/calm-mirrors-fail.md
@@ -1,0 +1,5 @@
+---
+'@shopify/eslint-plugin': major
+---
+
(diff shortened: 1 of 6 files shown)


BPScott:
You can comment `/snapit` on this PR to create a snapshot package release that you can then test within some repo to make sure it is good and you're happy with it.

mxstbr:
Awesomeee. Let's shipit!

PR #488, opened 2026-01-07
Update jsx-no-hardcoded-content to handle ternary and logical expressions

mateus:
This PR updates the `@shopify/jsx-no-hardcoded-content` ESLint rule to detect hardcoded strings within conditional (`ternary`) and logical expressions (`&&`, `||`). Previously, expressions like the following were not flagged:

```jsx
aria-label={showPassword ? 'Hide password' : 'Show password'}
```

## Changes

- Updated `isInvalidContent` to recursively check `ConditionalExpression` and `LogicalExpression`.

### DIFF
--- a/.changeset/smooth-cows-drum.md
+++ b/.changeset/smooth-cows-drum.md
@@ -1,0 +1,5 @@
+---
+'@shopify/eslint-plugin': minor
+---
+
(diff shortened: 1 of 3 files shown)


BPScott (approved):
Code changes seem good. We should fix the publishing to use OIDC before we merge this, so we can get the snapit back to working correctly

